PR #2, opened 2021-11-16
Add Makefile, Actions workflow, and Dangerfile

disposedtrolley:
This PR adds a lint and test pipeline with GitHub Actions and a Danger check, borrowing from previous work. I've also initialised a Go module and popped in an empty `main.go` so we can see the checks running.


jamestelfer (approved):
Nice work, LGTM

PR #4, opened 2021-11-17
Add flags package

disposedtrolley:
This PR introduces an experimental `flags` package for consuming feature flags from LaunchDarkly.

The package exposes a `Client` wrapping a subset of the LaunchDarkly functionality with a simplified configuration stage, and a `User` type allowing for user information to be supplied during flag evaluation in a standardised way.

The package can be used in two ways:

- An internally managed singleton, where consumers call package-level functions to configure, connect, and retrieve the singleton client to evaluate flags.
- A standard struct initialiser to create a client instance, where the caller is responsible for maintaining the instance as a singleton.

Example usage:

```go
func main() {
	proxyURL, _ := url.Parse("http://my.proxy.server")

	// Configure the singleton (proxy mode).
	_ = flags.Configure(
		flags.WithSDKKey("foo"),
		flags.WithInitWait(time.Second*2),
		flags.WithProxyMode(proxyURL),
	)

	// OR configure the singleton (daemon mode).
	_ = flags.Configure(
		flags.WithSDKKey("foo"),
		flags.WithInitWait(time.Second*2),
		flags.WithDaemonMode("my-dynamo-table-name", 10*time.Second),
	)

	// Establish the LaunchDarkly connection.
	_ = flags.Connect()

	// Retrieve the singleton instance and issue a query.
	c, _ := flags.GetDefaultClient()
	defer c.Shutdown()
	val, _ := c.QueryBool("my-flag-name", flags.NewAnonymousUser(), false)
}
```

```go
func main() {
	// Similar to above, but the consumer manages the singleton.
	proxyURL, _ := url.Parse("http://my.proxy.server")

	c, _ := flags.NewClient(
		flags.WithSDKKey("foo"),
		flags.WithInitWait(time.Second*2),
		flags.WithProxyMode(proxyURL),
	)

	_ = c.Connect()
	defer c.Shutdown()
	val, _ := c.QueryBool("my-flag-name", flags.NewAnonymousUser(), false)
}
```


disposedtrolley:
[Wrap SDK for Golang (ca-go/flags)](https://trello.com/c/XOxGYWlx/715-create-ca-go-flags)

x/flags/flags_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	t.Run("allows daemon mode to be configured", func(t *testing.T) {
+		err := flags.Configure(
+			flags.WithSDKKey("foobar"),
+			flags.WithDaemonMode("dynamo-table-name", 10*time.Second),
+		)
+		require.NoError(t, err)
+	})
+
+	t.Run("allows daemon mode to be configured with an alterate Dynamo base URL", func(t *testing.T) {
```
jamestelfer, line 59:
```suggestion
	t.Run("allows daemon mode to be configured with an alternate Dynamo base URL", func(t *testing.T) {
```

x/flags/flags.go
```diff
@@ -0,0 +1,198 @@
...
+	relayProxyURL string
+}
+
+type daemonModeConfig struct {
+	dynamoTableName string
+	dynamoBaseURL   string
+	cacheTTL        time.Duration
+}
+
+func NewClient(opts ...ConfigOption) (*Client, error) {
```
jamestelfer, line 117: If we expose the `Client` struct here, how would we introduce a different implementation do you think?

x/flags/flags.go
```diff
@@ -0,0 +1,198 @@
...
+
+func WithProxyMode(proxyURL *url.URL) ConfigOption {
+	return func(c *Client) {
+		c.proxyModeConfig = &proxyModeConfig{
+			relayProxyURL: proxyURL.String(),
+		}
+	}
+}
+
+func WithDaemonMode(dynamoTableName string, cacheTTL time.Duration) ConfigOption {
```
jamestelfer, line 73: These configuration options are basically LaunchDarkly. How worried are we about a different implementation?

x/flags/flags.go
```diff
@@ -0,0 +1,198 @@
...
+	if c.proxyModeConfig != nil {
+		c.wrappedConfig = configForProxyMode(c.proxyModeConfig)
+	} else if c.daemonModeConfig != nil {
+		c.wrappedConfig = configForDaemonMode(c.daemonModeConfig)
+	}
+
+	return c, nil
+}
+
+func (c *Client) Connect() error {
```
jamestelfer, line 144: WDYT about putting the `Client` struct and methods in a separate file to the config implementation?

x/flags/flags.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (c *Client) QueryString(key FlagName, user User, defaultValue string) (string, error) {
+	return c.wrappedClient.StringVariation(string(key), user.ldUser, defaultValue)
+}
+
+func (c *Client) QueryInt(key FlagName, user User, defaultValue int) (int, error) {
+	return c.wrappedClient.IntVariation(string(key), user.ldUser, defaultValue)
+}
+
+func (c *Client) RawClient() *ld.LDClient {
```
jamestelfer, line 192: If we were to change the implementation, this method would be a liability. Perhaps the return type should be `interface{}`.

I like the ethos of adding an escape hatch, but we should avoid branding it 😄

jamestelfer (changes requested):
Thanks for putting up a draft so quickly to get some conversation going.

At a high level, while functional, it does seem to leak a lot of "LaunchDarkly"ness through the config.

There's a couple of ways of looking at this. One is that if we were to have a different provider, we'd change (and then major version) the API.

Another way might be to expose a `flags.Client` interface, and keep the config in a different package. There are strengths and weaknesses in either case.

For a final version, we'd like to have a `doc.go` for the package docs, and some `Example` style tests showing the usage. I can see why you haven't done them yet though.

x/flags/flags.go
disposedtrolley: Yep that's a great idea -- I kept them both in the same file in this first pass as I was constantly referring between the two structs.

x/flags/flags.go
disposedtrolley: I agree. We can stipulate in the documentation that `RawClient()` should be used at your own risk, and the need to cast from an interface to `*ld.LDCLient` provides a nice hoop that users will need to jump through.

x/flags/flags.go
disposedtrolley: I was considering declaring `Client` as an interface (which is what I had in the [flags package for the PoC](https://github.com/cultureamp/employee-tasks-service/blob/4426310d0673b8b1998b8a47e1da9474b921583e/src/flags/flags.go#L119-L124) but it felt like another layer of indirection that doesn't provide much value.

I wanted to rely on Go's consumer-driven interfaces here so we can provide a slimmer library overall. Consumers of the library can use their own interface type that expose the portions of `Client` that they're interested in using. What do you think of this approach?

x/flags/flags.go
disposedtrolley: I'm not worried for the foreseeable future, but I definitely get your concern.

My initial goals with the wrappers wasn't to insulate LaunchDarkly's API surface from our services, but to provide a more convenient and standardised way of configuring and using their methods. We can probably reduce the coupling somewhat by abstracting the ideas of configuration and the client, but I'm worried that makes for a more confusing library and it feels like premature abstraction.

I sketched out what this package might look like if we provided more interfaces:

```go
type LDConfigOption func(c *LDClient)

type LDConfig struct {...}

func NewLDConfig(opts ...LDConfigOption) LDConfig {...}

type Client interface {
    Query(...)
    Shutdown()
    Connect()
}

type LDClient struct {...}

func NewLDClient(cfg LDConfig) Client {...}

func (c *LDClient) Query(...) {...}
func (c *LDClient) Shutdown() {...}
func (c *LDClient) Connect() {...}


func main() {
    config := NewLDConfig(...)
    client := NewLDClient(config)

    client.Query(...)
}
```

disposedtrolley:
Once we're happy with the API and implementation, I'll fix up the commit history.

x/launchdarkly/flags/user.go
```diff
@@ -0,0 +1,96 @@
...
+	userAttributeRealUserAggregateID = "realUserAggregateID"
+)
+
+// User wraps the LaunchDarkly user object.
+type User struct {
+	effectiveUserAggregateID string
+	realUserAggregateID      string
+	accountAggregateID       string
+
+	ldUser lduser.User
```
hallaji, line 24: I wonder whether we plan to expose entire LD user object or not? How about destructing the object here one by one and adding those that will be needed? That's gonna hide LD user struct as well.

hallaji (commented):
I think we need to come up with a consistent interface across different wrappers, e.g. what sort of user object/interface are we exposing to developers? It could be documented somewhere as a reference. What do you think?

x/launchdarkly/flags/user.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/cultureamp/ca-go/x/request"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+)
+
+const (
+	anonymousUser                    = "ANONYMOUS_USER"
+	userAttributeAccountAggregateID  = "accountAggregateID"
+	userAttributeRealUserAggregateID = "realUserAggregateID"
```
melm73, line 15: Same comment as the other wrappers, can we please not use `aggregate` in our field names. Within the monoliths we have the problem of multiple ids for these domains, but outside the monolith there is no need to make this distinction. There is only 1 id that we care about for accounts and users.

The JWT is a good example of this. No `aggregate` in sight.

(Side note... accounts are not even event sourced, so actually it isn't an aggregate id at all 😜 )

melm73 (dismissed):

disposedtrolley:
> I think we need to come up with a consistent interface across different wrappers,

Yes that's a great idea. We should keep the interface loose so it can be adapted to suit each language, but the overall concepts and usage patterns should be similar.

What I've been doing for the Ruby and Go wrappers is offering two ways of configuring and instantiating the client:

- As an internally managed singleton where the application can configure some globals during initialisation, and retrieve the singleton instance from the package at the time of use.
- A regular constructor that returns an instance of the client, that the application must manage itself.

Both approaches have two stages:

1. Configuration: options are specified in some manner and passed into the appropriate function depending on the approach above.
2. Connection: a connection to LaunchDarkly is established.
3. Usage: methods on the client instance are called (or the singleton is retrieved and methods on it are called).
4. Shutdown: the client provides a Shutdown method that closes connections to LaunchDarkly.

x/launchdarkly/flags/user.go
disposedtrolley: The LD user and client objects are provided as an escape hatch as we're not wrapping the entirety of the interface.

x/launchdarkly/flags/user.go
disposedtrolley: Yeah I've been meaning to change these, thanks for reminding me!

aimeegee (approved):
Great work James. I'll take yours as a model when adding other packages in

PR #6, opened 2021-11-25
feat: add request package

disposedtrolley:
This PR adds a `request` package exposing types and helper functions to create, add, and retrieve request-scoped attributes to context.Context.

There are two sets of attributes that can be unique per request:

- Request identifiers like the trace, correlation, and request IDs.
- User identifiers like their account and user IDs, for authenticated requests.

Both sets may need to be present in the context for certain request middleware to function correctly. An example is the Sentry error reporter, which is provided request and user identifiers where available.

### DIFF
--- a/x/request/requestids.go
+++ b/x/request/requestids.go
@@ -12,17 +12,22 @@ type RequestIDs struct {
 	CorrelationID string
 }
 
-// AddRequestIDsToContext returns a new context with the given RequestIDs
+// ContextWithRequestIDs returns a new context with the given RequestIDs
 // embedded as a value.
 func ContextWithRequestIDs(ctx context.Context, fields RequestIDs) context.Context {
 	return context.WithValue(ctx, requestIDsKey, fields)
 }
 
+// RequestIDsFromContext attempts to retrieve a RequestIDs struct from the given
+// context, returning a RequestIDs struct along with a boolean signalling
+// whether the retrieval was successful.
 func RequestIDsFromContext(ctx context.Context) (RequestIDs, bool) {
 	ids, ok := ctx.Value(requestIDsKey).(RequestIDs)
 	return ids, ok
 }
 
+// ContextHasRequestIDs returns whether the given context contains a RequestIDs
+// value.
 func ContextHasRequestIDs(ctx context.Context) bool {
 	_, ok := RequestIDsFromContext(ctx)
 	return ok
--- a/x/request/requestids.go
+++ b/x/request/requestids.go
@@ -1,0 +1,29 @@
+package request
+
+import "context"
+
+type contextValueKey string
+
+const requestIDsKey = contextValueKey("fields")
+
+// RequestIDs represent the set of unique identifiers for a request.
+type RequestIDs struct {
+	RequestID     string
+	CorrelationID string
+}
+
+// AddRequestIDsToContext returns a new context with the given RequestIDs
+// embedded as a value.
+func ContextWithRequestIDs(ctx context.Context, fields RequestIDs) context.Context {
+	return context.WithValue(ctx, requestIDsKey, fields)
+}
+
+func RequestIDsFromContext(ctx context.Context) (RequestIDs, bool) {
+	ids, ok := ctx.Value(requestIDsKey).(RequestIDs)
+	return ids, ok
+}
+
+func ContextHasRequestIDs(ctx context.Context) bool {
+	_, ok := RequestIDsFromContext(ctx)
+	return ok
+}
--- a/x/request/requestids_test.go
+++ b/x/request/requestids_test.go
@@ -2,6 +2,7 @@ package request_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/cultureamp/ca-go/x/request"
@@ -26,11 +27,44 @@ func TestContextWithRequestIDs(t *testing.T) {
 	assert.True(t, ok)
 }
 
+func ExampleContextWithRequestIDs() {
+	requestIDs := request.RequestIDs{
+		RequestID:     "123",
+		CorrelationID: "456",
+	}
+	ctx := context.Background()
+
+	ctx = request.ContextWithRequestIDs(ctx, requestIDs)
+
+	if requestIDsFromContext, ok := request.RequestIDsFromContext(ctx); ok {
+		fmt.Println(requestIDsFromContext.RequestID)
+		fmt.Println(requestIDsFromContext.CorrelationID)
+
+		// Output:
+		// 123
+		// 456
+	}
+}
+
 func TestRequestIDsFromContextMissing(t *testing.T) {
 	ctx := context.Background()
 
 	_, ok := request.RequestIDsFromContext(ctx)
 	assert.False(t, ok)
+}
+
+func ExampleContextHasRequestIDs() {
+	requestIDs := request.RequestIDs{
+		RequestID:     "123",
+		CorrelationID: "456",
+	}
+	ctx := context.Background()
+
+	ctx = request.ContextWithRequestIDs(ctx, requestIDs)
+
+	ok := request.ContextHasRequestIDs(ctx)
+	fmt.Println(ok)
+	// Output: true
 }
 
 func TestContextHasRequestIDsSucceeds(t *testing.T) {
--- a/x/request/requestids_test.go
+++ b/x/request/requestids_test.go
@@ -1,0 +1,48 @@
+package request_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/cultureamp/ca-go/x/request"
+	"github.com/stretchr/testify/assert"
+)
+
+func newRequestIDs() request.RequestIDs {
+	return request.RequestIDs{
+		RequestID:     "123",
+		CorrelationID: "456",
+	}
+}
+
+func TestContextWithRequestIDs(t *testing.T) {
+	ids := newRequestIDs()
+	ctx := context.Background()
+
+	ctx = request.ContextWithRequestIDs(ctx, ids)
+	idsFromContext, ok := request.RequestIDsFromContext(ctx)
+
+	assert.Equal(t, ids, idsFromContext)
+	assert.True(t, ok)
+}
+
+func TestRequestIDsFromContextMissing(t *testing.T) {
+	ctx := context.Background()
+
+	_, ok := request.RequestIDsFromContext(ctx)
+	assert.False(t, ok)
+}
+
+func TestContextHasRequestIDsSucceeds(t *testing.T) {
+	ctx := request.ContextWithRequestIDs(context.Background(), newRequestIDs())
+
+	ok := request.ContextHasRequestIDs(ctx)
+	assert.True(t, ok)
+}
+
+func TestContextHasRequestIDsFails(t *testing.T) {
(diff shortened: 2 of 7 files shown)


request/requestids.go
```diff
@@ -0,0 +1,46 @@
...
+import "context"
+
+type contextValueKey string
+
+const requestIDsKey = contextValueKey("fields")
+
+// RequestIDs represent the set of unique identifiers for a request.
+// JSON tags are included for log serialisation purposes.
+type RequestIDs struct {
+	TraceID       string `json:"trace_id"`
```
jamestelfer, line 12: What is the trace ID tracking? Could these be used in a context like an event (i.e. is the request nomenclature correct)? Are JSON tags really required here?

disposedtrolley: @melm73 and I explored several options for naming the package, and `request` seemed the most natural. The original implementation was named `context`, and users were aliasing the import to `gcontext` which didn't seem ideal.

1. I think this is populated with the AWS XRay ID by request middleware.
2. Yeah they could. We tossed up between a few names and "request" seems to cover the breadth of architectures we might use this for.
3. I took them out but popped them back as I thought they might be used for logging. Maybe not...

request/requestids.go
```diff
@@ -0,0 +1,46 @@
...
+	RequestID     string `json:"request_id"`
+	CorrelationID string `json:"correlation_id"`
+}
+
+// NewRequestIDs returns a new RequestIDs with the given request
+// identifiers.
+func NewRequestIDs(traceID string, requestID string, correlationID string) RequestIDs {
```
jamestelfer, line 19: Do we need a constructor method for this struct?

disposedtrolley: I reckon we should have one to disambiguate whether pointer or value semantics should be used for `RequestIDs`.

request/requestids.go
jamestelfer: (1) I'm not certain this still makes sense. Let's take it out and add it back if needed.
(2) As long as it's considered, I'm happy.
(3) Let's take them out and add them back if they're required.

request/user_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/cultureamp/ca-go/request"
+	"github.com/stretchr/testify/assert"
+)
+
+func newAuthenticatedUser() request.AuthenticatedUser {
+	return request.NewAuthenticatedUser("123", "456", "789")
+}
+
+func TestContextWithAuthenticatedUser(t *testing.T) {
```
jamestelfer, line 15: Let's add some `Example*` methods in

request/doc.go
```diff
@@ -0,0 +1,7 @@
+// package request exposes types and helper methods to create, add, and retrieve
+// request-scoped attributes to context.Context.
+//
+// Request-scoped attributes include identifiers like the request, trace, and
+// correlation IDs.When the request is authenticated, user identifiers like the
+// account and user aggregate IDs can also be added to the context.
```
jamestelfer, line 6:
```suggestion
// The request package exposes types and helper methods to create, add, and retrieve
// request-scoped attributes to context.Context.
//
// Request-scoped attributes include identifiers like the request, trace, and
// correlation IDs. When the request is authenticated, user identifiers like the
// account and user aggregate IDs can also be added to the context.
```

request/requestids_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/cultureamp/ca-go/request"
+	"github.com/stretchr/testify/assert"
+)
+
+func newRequestIDs() request.RequestIDs {
+	return request.NewRequestIDs("123", "456", "789")
+}
+
+func TestContextWithRequestIDs(t *testing.T) {
```
jamestelfer, line 15: Examples?

request/user.go
```diff
@@ -0,0 +1,47 @@
+package request
+
+import "context"
+
+const authenticatedUserKey = contextValueKey("payload")
```
jamestelfer, line 5:
```suggestion
const authenticatedUserKey = contextValueKey("authenticatedUser")
```

jamestelfer (changes requested):
Minor changes, but a good and clean first cut.

Let's add some examples for docs and get this in, and add the middleware later.

Question: `/x` prefix? I suppose this is currently pre-v1, so we might be OK?

request/doc.go
disposedtrolley: I've fixed the missing space, but I'm inclined to keep `Package request...` at the beginning of the doc. This appears to be the Go convention for package docs:

- https://pkg.go.dev/net/http#pkg-overview
- https://pkg.go.dev/encoding/gob#pkg-overview
- https://pkg.go.dev/github.com/stretchr/testify/assert?utm_source=godoc#pkg-overview

What are your thoughts?

disposedtrolley:
> Question: `/x` prefix? I suppose this is currently pre-v1, so we might be OK?

Yeah I think we should nest this under `/x` for now. It can graduate once we've migrated one or two applications to use this package.

request/requestids.go
jamestelfer: What benefit do we gain from this unless there are methods?

request/doc.go
jamestelfer: TIL, love the response. Keep it, and perhaps note it in the design.md?

x/request/user.go
```diff
@@ -0,0 +1,47 @@
...
+import "context"
+
+const authenticatedUserKey = contextValueKey("authenticatedUser")
+
+// AuthenticatedUser holds the identifiers of a user making an authenticated
+// request.
+type AuthenticatedUser struct {
+	AccountAggregateID       string
+	RealUserAggregateID      string
+	EffectiveUserAggregateID string
```
jamestelfer, line 12:
```suggestion
	CustomerAccountID       string
	UserID string
	RealUserID      string
```

I don't think adding `Aggregate` in here is helpful. For JWT auth, that's what they're going to be set to, but I'm not convinced it should be part of the field name.

x/request/user.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// NewAuthenticatedRequest returns a new AuthenticatedRequest with the given
+// IDs.
+func NewAuthenticatedUser(accountAggregateID, realUserAggregateID, effectiveUserAggregateID string) AuthenticatedUser {
```
jamestelfer, line 17: Same here regarding the constructor.

x/request/requestids.go
```diff
@@ -0,0 +1,43 @@
...
+
+// RequestIDs represent the set of unique identifiers for a request.
+type RequestIDs struct {
+	RequestID     string
+	CorrelationID string
+}
+
+// NewRequestIDs returns a new RequestIDs with the given request
+// identifiers.
+func NewRequestIDs(requestID string, correlationID string) RequestIDs {
```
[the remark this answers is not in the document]
jamestelfer: To add to this, the two IDs are strings, and without named parameters this code becomes more prone to errors. I know Goland puts names in, but that's an IDE specific helper that kind of exposes the issue.

x/request/requestids_test.go
```diff
@@ -0,0 +1,62 @@
...
+	ctx := context.Background()
+
+	ctx = request.ContextWithRequestIDs(ctx, ids)
+	idsFromContext, ok := request.RequestIDsFromContext(ctx)
+
+	assert.Equal(t, ids, idsFromContext)
+	assert.True(t, ok)
+}
+
+func ExampleContextWithRequestIDs() {
```
jamestelfer, line 27: Love examples

x/request/user_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func TestAuthenticatedUserFromContextMissing(t *testing.T) {
+	ctx := context.Background()
+
+	_, ok := request.AuthenticatedUserFromContext(ctx)
+	assert.False(t, ok)
+}
+
+func TestContextHasAuthenticatedUserSuceeds(t *testing.T) {
```
jamestelfer, line 52: Seems like a method that could have an example too?

jamestelfer (changes requested):
A couple of points of discomfort in a generally awesome PR>

x/request/requestids.go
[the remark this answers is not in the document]
disposedtrolley: > I'd prefer to avoid constructor methods unless necessary.

Since this is serving as a bag of data with no methods I'm happy to remove it.

For more complex structs with methods (that can have pointer _or_ value receivers), it can be useful to denote its usage semantics with a constructor. If the constructor returns a pointer, the reader can expect a degree of mutability for instance.

x/request/user.go
disposedtrolley: I'm happy with the shorter names; I wasn't sure how useful being explicit with `aggregate` would be.

x/request/user_test.go
disposedtrolley: Yup could be useful. I'll add the equivalent for RequestIDs too.

x/request/user.go
disposedtrolley: Would it make sense to use the same names as web gateway? https://github.com/cultureamp/web-gateway/blob/ccf9db2fabb2a798a9905475a3fce63fdc192811/identity/identity.go#L20-L24

jamestelfer: That is what's exposed in the JWT. I'd follow that _except_ for `EffectiveUserID` becomes `UserID`. Every single time I see `Effective` and `Real` I have to ask myself which one I'm going to use. Even `Real` is ambiguous.

`UserID` and `ImpersonatingUserID` might be better.

x/request/requestids.go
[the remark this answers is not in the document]
jamestelfer: Note that a struct should generally have only one or the other, and the receiver type of any methods show the expected use.

x/request/user.go
disposedtrolley: Yeah, `Real` is only real in context. Would `ImpersonatingUserID` be the ID of the person who is impersonating, or being impersonated?

Naming is hard.

disposedtrolley: I'm also happy to keep it as `AccountID`, `UserID`, and `RealUserID` to be consistent with other systems for now.

x/request/requestids.go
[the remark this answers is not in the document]
disposedtrolley: True, but I think a constructor makes that more explicit.

x/request/user.go
```diff
@@ -0,0 +1,37 @@
...
+import "context"
+
+const authenticatedUserKey = contextValueKey("authenticatedUser")
+
+// AuthenticatedUser holds the identifiers of a user making an authenticated
+// request.
+type AuthenticatedUser struct {
+	CustomerAccountID string
+	UserID            string
+	RealUserID        string
```
jamestelfer, line 12:
```suggestion
	// CustomerAccountID is the ID of the currently logged in user's parent account/organization, sometimes known as the "account_aggregate_id".
	CustomerAccountID string
	// UserID is the ID of the currently authenticated user, and will generally be a "user_aggregate_id".
	UserID            string
	// RealUserID, when supplied, is the ID of the user who is currently impersonating the current "UserID".
	// This value is optional.
	RealUserID        string
```

jamestelfer (approved):
Lets go!

PR #7, opened 2021-11-26
feat: add Sentry package

disposedtrolley:
This PR adds an `errorreport` package to configure and report errors to Sentry.

- Configuration is global to the application instance; calling `Configure` during application initialisation configures Sentry until the application shuts down.
- Errors can be reported ad-hoc by calling `ReportError` directly.
- HTTP and Goa endpoint middleware can be instantiated to automatically detect and report errors.


go.mod
```diff
@@ -5,7 +5,8 @@ go 1.17
 require github.com/stretchr/testify v1.7.0
 
 require (
-	github.com/davecgh/go-spew v1.1.0 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/getsentry/sentry-go v0.11.0
```
hughevans, line 9: You might like to add excludes for all the frameworks that `getsentry/sentry-go` pulls in because we know there’s CVEs in some of those. eg. https://github.com/cultureamp/glamplify/pull/39/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R37-R42

hughevans: More info here https://github.com/getsentry/sentry-go/issues/156

disposedtrolley: Thanks Hugh, I'll have a look!

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,56 @@
...
+	"goa.design/goa"
+)
+
+type OnRequestPanicHandler func(context.Context, http.ResponseWriter, error)
+
+func NewHTTTPMiddleware(onRequestPanic OnRequestPanicHandler) func(http.Handler) http.Handler {
+	sentryWrapper := sentryhttp.New(sentryhttp.Options{
+		// Repanic to propagate the error to the onRequestPanic handler.
+		Repanic: true,
+	})
```
disposedtrolley, line 19: @jamestelfer do we need to append the user and request attributes to the Sentry scope here too? I noticed it was being done explicitly for the Goa middleware which calls our `ReportError` function.

x/sentry/errorreport/middleware_test.go
```diff
@@ -0,0 +1,132 @@
+package errorreport_test
```
disposedtrolley, line 1: I'm not sure if we need example tests here. There'd be a fair bit of boilerplate in the setup, and I'm unsure what we'd use for the output.

x/sentry/errorreport/middleware_test.go
```diff
@@ -0,0 +1,132 @@
...
+		// ...recovers the panic...
+		// nolint:bodyclose
+		assert.Equal(t, http.StatusTeapot, w.Result().StatusCode)
+
+		// ...and reports the error to Sentry.
+		assert.Len(t, mockSentryTransport.Events(), 1)
+	})
+}
+
+func TestGoaEndpointMiddleware(t *testing.T) {
```
disposedtrolley, line 94: I know this isn't doing much, but I wanted to make sure our middleware is actually executing the wrapped code and attempting to send errors where necessary.

disposedtrolley:
[Create ca-go/sentry](https://trello.com/c/AFDV26ZB/723-create-ca-go-sentry)

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,60 @@
...
+
+	sentryhttp "github.com/getsentry/sentry-go/http"
+	"goa.design/goa"
+)
+
+type OnRequestPanicHandler func(context.Context, http.ResponseWriter, error)
+
+// NewHTTPMiddleware returns an http.Handler that reports any errors to Sentry,
+// recovers from panics, and calls the given OnRequestPanic handler as necessary.
+func NewHTTTPMiddleware(onRequestPanic OnRequestPanicHandler) func(http.Handler) http.Handler {
```
melm73, line 17: Typo:  `NewHTTTPMiddleware` has an extra `T`

x/request/requestids.go
```diff
@@ -0,0 +1,43 @@
+package request
+
+import "context"
+
+type contextValueKey string
+
+const requestIDsKey = contextValueKey("fields")
+
+// RequestIDs represent the set of unique identifiers for a request.
+type RequestIDs struct {
```
melm73, line 10: Nice naming!! 😉

melm73 (commented):
Looks good to me, looks like you're still waiting on answers to your questions, so I won't approve just yet.

x/sentry/errorreport/middleware.go
disposedtrolley: https://www.youtube.com/watch?v=3m6Blqs0IgY

x/sentry/errorreport/middleware_test.go
```diff
@@ -0,0 +1,132 @@
...
+		endpoint := func(ctx context.Context, req interface{}) (interface{}, error) {
+			endpointCalled = true
+
+			return nil, errors.New("boom")
+		}
+
+		mw := errorreport.NewGoaEndpointMiddleware()
+
+		sut := mw(endpoint)
+		_, _ = sut(context.Background(), nil)
```
jamestelfer, line 127: Anything to assert on these returns?

x/sentry/errorreport/middleware_test.go
```diff
@@ -0,0 +1,132 @@
...
+		endpoint := func(ctx context.Context, req interface{}) (interface{}, error) {
+			endpointCalled = true
+
+			return nil, nil
+		}
+
+		mw := errorreport.NewGoaEndpointMiddleware()
+
+		sut := mw(endpoint)
+		_, _ = sut(context.Background(), nil)
```
jamestelfer, line 108: And here?

x/sentry/errorreport/middleware_test.go
jamestelfer: entirely appropriate

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,60 @@
...
+
+	sentryhttp "github.com/getsentry/sentry-go/http"
+	"goa.design/goa"
+)
+
+type OnRequestPanicHandler func(context.Context, http.ResponseWriter, error)
+
+// NewHTTPMiddleware returns an http.Handler that reports any errors to Sentry,
+// recovers from panics, and calls the given OnRequestPanic handler as necessary.
+func NewHTTPMiddleware(onRequestPanic OnRequestPanicHandler) func(http.Handler) http.Handler {
```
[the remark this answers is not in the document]
jamestelfer: Also, if we implemented this in another context (like a queue), what would be needed? This shouldn't block us merging this PR

jamestelfer (changes requested):
Most of the comments aren't that high impact IMO. Understanding the required use and getting that in the doco is the biggest thing.

x/sentry/errorreport/errorreport.go
```diff
@@ -0,0 +1,113 @@
...
+
+	sentryConfig = cfg
+
+	return nil
+}
+
+// Connect initialises the Sentry client. An error is returned if the
+// client is not yet configured, or an initialisation error occurs.
+func Connect() error {
+	if sentryConfig == nil {
```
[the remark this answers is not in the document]
disposedtrolley: I've updated this to return early if `connected` is already true.

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,60 @@
...
+	sentryWrapper := sentryhttp.New(sentryhttp.Options{
+		// Repanic to propagate the error to the onRequestPanic handler.
+		Repanic: true,
+	})
+
+	return func(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
disposedtrolley: Yep, I'll adjust the existing comments to be a bit clearer.

x/sentry/errorreport/middleware.go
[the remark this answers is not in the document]
disposedtrolley: A good default handler might write a 500 status code with an empty body. Should we include something like this?

```go
func DefaultOnRequestPanicHandler(ctx context.Context, w http.ResponseWriter, err error) {
	w.WriteHeader(http.StatusInternalServerError)
}
```

x/sentry/errorreport/middleware.go
disposedtrolley: I've added request fields to the Sentry scope in the HTTP middleware too. I've also improved the tests so we augment the context with sample values, and check the values have been propagated in the Sentry error report.

x/sentry/errorreport/config.go
```diff
@@ -0,0 +1,96 @@
...
+
+	sentryOpts sentry.ClientOptions
+	connected  bool
+}
+
+// Option is a function type that can be provided to Configure to modify the
+// behaviour of Sentry.
+type Option func(c *config)
+
+type sentryBeforeFilter func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event
```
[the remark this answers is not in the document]
disposedtrolley: No reason, I think it was private when we were drafting the API and I didn't check to change it.

x/sentry/errorreport/config.go
```diff
@@ -0,0 +1,96 @@
...
+
+// WithEnvironment configures Sentry for the given environment, e.g. production-us.
+// This is a mandatory option.
```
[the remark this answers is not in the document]
disposedtrolley: Much better, thanks!

x/sentry/errorreport/middleware.go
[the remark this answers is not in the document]
jamestelfer: I'd say write a JSON API error by default, with the message "Internal server error" and set the mime type to `application/json`. This is more likely to match what the service is exposing in its API interface

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,70 @@
...
+func NewHTTPMiddleware(onRequestPanic OnRequestPanicHandler) func(http.Handler) http.Handler {
+	sentryWrapper := sentryhttp.New(sentryhttp.Options{
+		// Repanic to propagate the error to the onRequestPanic handler.
+		Repanic: true,
+	})
+
+	// Returns a handler that repanics any panics that Sentry swallows by default. The
+	// new panic is recovered to enable the onRequestPanic function to be executed so
+	// it can perform any further manipulation of the HTTP response before it's sent
+	// to the client.
```
jamestelfer, line 29:
```suggestion
	// Returns a handler that configures Sentry to report the panic, and then re-panic.
	//
	// The new panic is recovered to enable the onRequestPanic function to be executed so
	// it can perform further manipulation of the HTTP response before it's sent
	// to the client.
	//
	// The most common action on panic will be to change the HTTP response code
	// and request body to reflect an error occurring.
```

x/request/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package request exposes types and helper methods to create, add, and retrieve
+// request-scoped attributes to context.Context.
+//
+// Request-scoped attributes include identifiers like the request and correlation
+// IDs. When the request is authenticated, user identifiers like the account and
+// user aggregate IDs can also be added to the context.
+package request
```
disposedtrolley, line 7: Document usage here for configuration and both middlewares.

x/sentry/errorreport/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package errorreport enables you to configure Sentry for error reporting.
+// A general ReportError function is provided for ad-hoc reporting, as well
+// as several options for middleware. These middleware detect and report
+// errors automatically.
+package errorreport
```
disposedtrolley, line 5: Document usage here for configuration and both middlewares.

x/sentry/errorreport/errorreport.go
```diff
@@ -0,0 +1,128 @@
...
+	cfg.sentryOpts = sentryOpts
+
+	sentryConfig = cfg
+
+	return nil
+}
+
+// Connect initialises the Sentry client. An error is returned if the
+// client is not yet configured, or an initialisation error occurs.
+func Connect() error {
```
disposedtrolley, line 69: Collapse this and `Configure` into a single `Init` function.

x/sentry/errorreport/middleware.go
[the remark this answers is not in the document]
disposedtrolley: I've added a default implementation but have run across some issues with how the panic handler constructs the error response.

1. If the request handler sets the status code prior to panicking, there is no way for the panic handler to re-write the status code to 500.
2. If the request handler writes to the response body prior to panicking, the panic handler will append the JSON:API error response to whatever was written already.

x/sentry/errorreport/doc.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// HTTP middleware can be used. Passing in nil uses the default panic
+// handler. See the OnRequestPanicHandler type if you wish to supply
+// your own.
+//   mw := middleware.NewHTTPMiddleware(nil)
+//   mw(myHTTPHandler)
+//
+// Goa middleware can be used:
+//   mw := errorreport.NewGoaMiddleware()
+//   mw(myGoaEndpoint)
```
jamestelfer, line 26:
```suggestion
// Goa middleware can be used:
//   mw := errorreport.NewGoaMiddleware()
//   mw(myGoaEndpoint)
//
// This is recommended when using Goa, as it offers reporting of all errors returned from 
// the generated logic types.
```

x/sentry/errorreport/doc.go
```diff
@@ -0,0 +1,27 @@
...
+// errors automatically.
+//
+// You configure and initialise errorreport using Init():
+//   err := errorreport.Init(
+//			  errorreport.WithDSN("foo")
+//			  errorreport.WithRelease("bar")
+//			  errorreport.WithEnvironment("baz"))
+//   if err != nil {
+//     // handle initialisation error
+//	 }
```
jamestelfer, line 13:
```suggestion
// if err != nil {
//    // handle initialisation error
// }
```

x/sentry/errorreport/middleware.go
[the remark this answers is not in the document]
jamestelfer: There are ways to get around both of these, at least in part. For a failed request though, I'm happy with a broken response format -- either way it's broken, adding the additional JSON is the least of their worries.

It's worth noting too that most `panic`-worthy behaviour comes from logic that occurs before writing a response. This means that while the above can occur, the default behaviour that tries to return a reasonable response works better than the default of not doing anything.

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,105 @@
...
+
+type jsonAPIErrorResponse struct {
+	Errors []jsonAPIError `json:"errors"`
+}
+
+type jsonAPIError struct {
+	Status string `json:"status"`
+	Title  string `json:"title"`
+}
+
```
jamestelfer, line 27: Note that we'll get these shortly with the API errors package.

x/sentry/errorreport/middleware.go
```diff
@@ -0,0 +1,105 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusInternalServerError)
+	_ = json.NewEncoder(w).Encode(jsonAPIErrorResponse{
+		Errors: []jsonAPIError{
+			{
+				Status: "500",
+				Title:  "Internal Server Error",
+			},
+		},
+	})
```
jamestelfer, line 40: Note that given that there is nothing dynamic in this response, a literal JSON string would work just as well.

jamestelfer (changes requested):
Minor doco nits, and a note that you're two methods off 100% coverage ;)

x/sentry/errorreport/middleware.go
disposedtrolley: Yeah I'm inclined to use a literal JSON string instead, until the API errors package arrives in theatres.

x/sentry/errorreport/middleware.go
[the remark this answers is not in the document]
disposedtrolley: Yep that's what I was thinking too. We should be okay for the majority of cases.

jamestelfer (approved):
:hype_frog: Let's go!

PR #10, opened 2022-02-02
feat: prefix evaluation context keys

disposedtrolley:
This PR modifies the `flags` package to prefix the keys used in the evaluation context sent to LaunchDarkly. The prefix will be the type of entity represented by the context, for example a `user` or `survey`, and is used to logically distinguish different business entities used in LD for flag targeting.

Along with this change, the PR updates the docs for the `flags` package to fix some whitespace problems. The PR also adds package documentation to `evaluationcontext` as it's a bit of a tricky concept to explain.

### DIFF
--- a/x/launchdarkly/flags/evaluationcontext/doc.go
+++ b/x/launchdarkly/flags/evaluationcontext/doc.go
@@ -1,0 +1,25 @@
+// Package evaluationcontext defines the kinds of evaluation contexts you can
+// provide in a query for a flag or product toggle. Constructor functions are
+// exposed to create valid instances of evaluation contexts.
+//
+// An evaluation context is simply a bag of attributes keyed by a unique
+// identifier. These values are used in two situations:
+//   1. When creating a flag or segment, the attributes are used to form the
+//      targeting rules. For example, "if the user's accountID is 123, return
+//      false for this flag".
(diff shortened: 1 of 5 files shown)


disposedtrolley:
[Modify constructor for evaluation context to add entity prefix to user key](https://trello.com/c/7OgTfv65/847-modify-constructor-for-evaluation-context-to-add-entity-prefix-to-user-key)

x/launchdarkly/flags/evaluationcontext/doc.go
```diff
@@ -0,0 +1,25 @@
+// Package evaluationcontext defines the kinds of evaluation contexts you can
```
hughevans, line 1: Small typo

hughevans (approved):
The tests in this package are great @disposedtrolley.  Really useful to have some nice canonical examples of good go tests.

disposedtrolley:
> update: having said that there are more public functions in user.go that should have tests ;)

Do you mean client.go? If so, yeah, and I'm not very proud of this haha.

One of the next steps for the wrapper is to provide an easy way for teams to use mocks in place of a real connection to LD. I'll need to figure out a way to dependency inject the LD client, which will make it easier to unit test the query functions in client.go

PR #16, opened 2022-03-22
feat: add AWS Lambda integration with Sentry

aimeegee:
## Purpose

Add support for Sentry reporting middleware similar to that currently supplied for `http.Handler`s.

There are a couple of complications here: 

1. The Sentry Go SDK doesn't have same direct support for AWS Lambda integration like the HTTP middleware.
2. There are a wide range of function signatures that Lambda supports.

For (1), we've brought in an approach inspired by other middleware available in the Sentry SDK; this was a mere pothole in the way of progress.

For (2), it's more complicated. If we supported `func (context.Context, interface{}) error`, the Lambda runtime deserializes the payload to a `map[string]any`, complicating its usage in client code. If we supported multiple specific signatures (e.g. `func(context.Context, events.KinesisEvent) error`), then we would have around 9 separate signatures to support.

In the end we went with generics, allowing support for `func(context.Context, TIn) error` and `func(context.Context, TIn) (TOut, error)`. This covers the highlights of the [allowed function signatures](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html). There is an opinion here: that the function should take a context and a typed input, and return an error, but it really doesn't place much restriction on usage.

It's noteworthy that Go 1.18 with generics has not long been released, so we should bear this in mind as we integrate it with our projects. The choice to take this approach was judged the risk of a runtime bug in the short term as fairly low, balanced against the need to support a more complex approach for a longer period.

I'd expect to take a similar approach with Datadog when we bring that in.

Note: [there is an example](https://github.com/cultureamp/ca-go/pull/16/files#diff-a3887c6bd5dceeee1e1fc2e846b702da405b1a9c9dc3fd9462f87d699a42abd4R54) showing how this middleware is used.

## Context

We're in the process of adding Sentry support in an internal project that uses Kinesis, so it's the right time to do this.

Further reading: [Lambda function examples](https://github.com/aws/aws-lambda-go/blob/main/events/README.md) for different event types.


jamestelfer (commented):
As very general comments:

- we need to look at the names we've used here and make sure they make sense in this context. For example, we should look at `LambdaHandler` instead of `Handler`, and `Options` is probably too generic as well.
- the documentation on any public methods is really important for this repo
- Let's try to include a couple of `Example` tests where we can, as they become a really practical (and living) part of the documentation.

jamestelfer:
Useful doco: https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html

aimeegee:
[Add Sentry exception monitoring (via ca-go) to the FusionAuth hooks lambdas](https://trello.com/c/u0c0YzDw/927-add-sentry-exception-monitoring-via-ca-go-to-the-fusionauth-hooks-lambdas)

jamestelfer (approved):

x/sentry/errorreport/lambda.go
```diff
@@ -0,0 +1,97 @@
...
+// a Lambda function that has a payload type of TIn and returns the tuple TOut,error.
+func LambdaWithOutputMiddleware[TIn any, TOut any](nextHandler lambdafunction.HandlerWithOutputOf[TIn, TOut], config ...LambdaOption) lambdafunction.HandlerWithOutputOf[TIn, TOut] {
+	options := configure(config)
+
+	return func(ctx context.Context, event TIn) (TOut, error) {
+		defer beforeHandler(ctx, options)()
+
+		out, err := nextHandler(ctx, event)
+
+		afterHandler(ctx, err)
```
disposedtrolley, line 60: Do we need separate before and after handlers, or can we report the error in the beforeHandler?

Any context mutations performed by the nextHandler won't be propagated to the afterHandler.

x/sentry/errorreport/lambda_example_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Handler is the lambda handler function with the logic to be executed. In this
+// case, it's a Kinesis event handler, but this could be a handler for any
+// Lambda event.
+func Handler(ctx context.Context, event events.KinesisEvent) error {
+	for _, record := range event.Records {
+		err := processRecord(ctx, record)
+		if err != nil {
+			return err
+		}
```
disposedtrolley, line 68:
```suggestion
		if err := processRecord(ctx, record); err != nil {
			return err
		}
```

nitpick

disposedtrolley (changes requested):
Looks good! Generics seem pretty tame in Go, and very useful in this context.

I have some minor suggestions and a nitpick, but really happy with this change.

x/sentry/errorreport/lambda.go
aimeegee: We can't handle the error in beforeHandler like this until it's returned by nextHandler() right? So if we wanna merge beforeHandler and afterHandler, does that mean we gonna end up with:
```
out, err := nextHandler(ctx, event)
defer afterHandler(ctx, options, err) // doing error reporting and panic handling here
```
Would that cause any panic being missed...?

disposedtrolley: Ahh yeah, I'm not sure how I missed that!

disposedtrolley (approved):

x/sentry/errorreport/lambda.go
jamestelfer: Marking this as resolved: the current state is correct AFAICT

PR #19, opened 2022-04-04
feat: stop prefixing user keys in the evaluation context

disposedtrolley:
Modifies the LaunchDarkly wrapper.

Removes the `user.` prefix from the key of the User evaluation context.
This optimises for the common case -- targeting against user attributes
-- and reduces complexity that would be needed in systems wishing to
integrate with our model of a user.


disposedtrolley:
[Update wrappers to not namespace `key` for user entities](https://trello.com/c/A5ZiVbgM/1006-update-wrappers-to-not-namespace-key-for-user-entities)

PR #20, opened 2022-04-08
fix: don't route analytics events through the LD Relay

disposedtrolley:
Configures the proxy mode option of the LaunchDarkly SDK so that analytics events are _not_ routed through the Relay Proxy. This is being done because:

- Routing events through the Relay generates additional traffic.
- The [default Relay configuration](https://github.com/launchdarkly/ld-relay/blob/v6/docs/configuration.md#file-section-events) has event forwarding turned off.
- The use case for sending events via the Relay only highlights a benefit for [PHP environments](https://github.com/launchdarkly/ld-relay/blob/v6/docs/events.md).

### DIFF
--- a/x/launchdarkly/flags/config.go
+++ b/x/launchdarkly/flags/config.go
@@ -121,7 +121,7 @@ func WithDynamoBaseURL(baseURL *url.URL) ConfigOption {
 
 func configForProxyMode(cfg *proxyModeConfig) ld.Config {
 	return ld.Config{
-		ServiceEndpoints: ldcomponents.RelayProxyEndpoints(cfg.relayProxyURL),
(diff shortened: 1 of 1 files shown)


jamestelfer:
Please update the PR description with the "why" this change is being done.

disposedtrolley:
I've thought about it, and I think I'll enable event forwarding in the Relay rather than reconfiguring each wrapper. Routing all server-side traffic to LD via the Relay seems cleaner overall, with no downsides besides increased load on the Relay.

To answer your questions:

> what happens to the library and the relay if LD was offline? Would there be a delay when sending these events?

The events are buffered in memory (regardless of whether we connect directly or via the Relay) and sent as a batch. I'll need to find out what the retry mechanism is like.

> in the Lambda case, where it's talking directly to dynamo, does it still send these events? Are they synchronous or asynchronous?

It still sends events asynchronously in the same way as other server-side SDKs.

PR #22, opened 2022-04-26
feat: allow config overrides for Proxy and Lambda modes

disposedtrolley:
Modifies how the final configuration for the Proxy and Lambda modes are
constructed so the user is able to supply configuration overrides via
ConfigOption arguments to the NewClient() function. The "base"
configuration is sourced from the LAUNCHDARKLY_CONFIGURATION environment
variable and provides a sensible default for most users. Users with
specific needs can opt to force the client into a specific mode, and
supply relevant configuration overrides for those modes.

For example, the WithProxyMode() function allows the user to specify an
alternate URL for the Relay Proxy (different from the one we specify in
the environment variable). WithLambdaMode() allows the user to configure
the client to connect directly to a DynamoDB table, and customise the
cache TTL and Dynamo base URL as necessary.

### DIFF
--- a/x/launchdarkly/flags/client.go
+++ b/x/launchdarkly/flags/client.go
@@ -17,6 +17,10 @@ type Client struct {
 	mode          mode
 	wrappedConfig ld.Config
 	wrappedClient *ld.LDClient
+
+	// Optional config overrides.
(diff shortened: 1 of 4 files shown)


El-Wel (approved):
Easy to follow, looks great :)

PR #24, opened 2022-05-16
create wrapper for logrus

aimeegee:
## Purpose
Implemented a logging wrapper which enables you to do pretty logging using logrus. 

### What's included in the logger
- Support log with context, embed request IDs, user IDs, and environment config from the context value.
- Support different log levels with customise fields
- Compatible with splunk and datadog
- Output the log as pretty json format

### Why we decide to use `logrus`
- We started with defining how would the API look like, i.e. how we wanna call the logger: `logger.withField().Debug()`
- `logrus` supports that type of API as well as all the requirements listed above
- `logrus` is one of the most popular top ranking Go logging libs
- Then we compared using a 3rd party lib vs implement a logging lib according to the requirement by ourselves. Using `logrus` has much less effort according to what we need and the effort of maintainence.

### Reasons for not using any other lib
- No, we didn't compare them. `logrus` turns out to be easy to use, and fits all our need very well as we try it, so we just decided to go with it. 
- Could consider list several 3rd lib candidates that fits all the requirements above, then compare them to pick the one best suits us if `logrus` is proved to be not good enough in the future.

### What's not included

- Detailed error tracing provide by [glamplify](https://github.com/cultureamp/glamplify/blob/master/log/system.go).
  - The current `getErrorValues()` method requires the error we pass in should be a google error type or at least supports Stack interface.
  - The wrapper displays error location which contains the file path, method name and line number. We think it's good level of details for now. It would be easy to add a `WithErrorTrace()` method in `logger.go` if needed in the future.
- Use datadog as logging sink
  - This is going to be the next task. We found this [format requirement doc](https://docs.datadoghq.com/logs/log_collection/?tab=host) in order to make it work.
  - Better to check with SRE team about what should be the standard attributes we would like to included into datadog log as well.


hughevans:
Closed in favour of Zerolog #52

PR #25, opened 2022-05-18
Use coveralls.io for coverage

jamestelfer:
## Purpose

Introduce coveralls.io for coverage measurement and reporting for the repo, as a replacement for our home-grown code that wasn't really meeting what we needed.

This PR removes the existing implementation in favour of reporting to Coveralls. We will enable PR commenting at some point shortly: that's separate to the reporting side of things.

> Note that since the checks now run on `pull_request_target`, they won't run here until this PR is merged. The `_target` version is used to work better with OSS forks of the repo.

## Context

We're trying out the value of coveralls.io on our open source repos.

## Extra

The linters were complaining, possibly because we hadn't pinned the version in use. I've pinned it now and fixed the issues. We will need to update the version, but this should happen as a mindful step so we can fix issues or disable linters as needed.


.github/workflows/pipeline.yml
```diff
@@ -1,5 +1,12 @@
 name: pipeline
-on: [pull_request]
+on:
+  # use PR target to allow for forks
+  pull_request_target:
+
+  # run on push to main so coverage is generated
+  push:
+    branches:
+      - "main"
```
melm73, line 9: Do we intend to use renovate in this project? we've seen that running tests "on pull_request" means that renovate will consider all dependency changes to "pass" the build.

hughevans (approved):

.github/workflows/pipeline.yml
jamestelfer: You're right that this would be something to look at -- I'd have to look at how this could work properly in that case and still do the right thing with forks.

PR #28, opened 2022-05-24
Enable Coveralls comments

jamestelfer:
Triggers using the `pull_request` event for non-main branches. The `pull_request` event provides the PR URL as one of the event parameters, something Coveralls needs to be able to comment on the PR.

Also, the `COVERALLS_TOKEN` isn't required for the push to work: these are only required for private repos.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,6 +1,10 @@
 name: pipeline
 on:
   push:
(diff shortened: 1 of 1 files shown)


melm73 (approved):
Approving, but we'll need to revisit this if we want renovate to work in this repo.... how would both of those tools work?

PR #30, opened 2022-05-26
Allow arbitrary tags in error reports

jamestelfer:
Enables additional tags to be added as part of initial configuration. These tags will be included in every report, and can be used in Sentry for grouping and searching.

For example, you can add tags singly:

```go
errorreport.WithTag("animal", "gopher"),
```

Or many in one go:

```go
errorreport.WithTags(map[string]string{
	"genus":   "phoenicoparrus",
	"species": "jamesi",
}),
```

As a side effect, this change stops adding empty values for the build details fields (`farm`, `buildNumber`, `branch`, `commit`). The empty values were being added as a side-effect of using the library, regardless of the intent of the user. Not all processes will use `farm` (a Culture Amp concept), so an empty value will just be unwanted clutter.

### DIFF
--- a/x/sentry/errorreport/config.go
+++ b/x/sentry/errorreport/config.go
@@ -18,6 +18,8 @@ type config struct {
 	transport    sentry.Transport
 }
 
+// tag adds the specified name/value pair to the tags map, skipping any where
+// the key or the value is zero-length.
 func (c *config) tag(name string, value string) {
 	if len(name) > 0 && len(value) > 0 {
 		c.tags[name] = value
--- a/x/sentry/errorreport/config.go
+++ b/x/sentry/errorreport/config.go
@@ -12,13 +12,16 @@ type config struct {
 	release     string
 	debug       bool
 
-	buildNumber string
-	branch      string
-	commit      string
-	farm        string
+	tags map[string]string
 
 	beforeFilter SentryBeforeFilter
 	transport    sentry.Transport
+}
+
+func (c *config) tag(name string, value string) {
+	if len(name) > 0 && len(value) > 0 {
+		c.tags[name] = value
+	}
 }
 
 // Option is a function type that can be provided to Configure to modify the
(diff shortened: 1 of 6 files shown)


x/sentry/errorreport/errorreport.go
```diff
@@ -61,10 +63,9 @@ func Init(opts ...Option) error {
 	// Add build information to the scope for all error reports.
 	// This can't be done before we initialise the Sentry client.
 	sentry.ConfigureScope(func(scope *sentry.Scope) {
-		scope.SetTag("build_number", cfg.buildNumber)
-		scope.SetTag("branch", cfg.branch)
-		scope.SetTag("commit", cfg.commit)
-		scope.SetTag("farm", cfg.farm)
+		if cfg.tags != nil {
+			scope.SetTags(cfg.tags)
+		}
```
melm73, line 68: does this mean that our usage of the library in our lambdas changes? ie. instead of setting farm in the config, we need to instead setup buildnumber, branch, commit and farm as tags?

jamestelfer: No, this is purely internal to the configuration. The `WithBuildDetails` API remains unchanged.

The only change is that instead of called `Decorate` after the `Init` call and throwing away the callback, we can just add the extra details in the `Init` call directly.

If the farm etc isn't set though, we won't be sending an empty string for that tag. It just won't be present.

jamestelfer: Internally, the build details are just added directly to the `tags` map instead of being tracked separately.

x/sentry/errorreport/config.go
```diff
@@ -90,9 +93,31 @@ func WithServerlessTransport() Option {
 // error reports.
 func WithBuildDetails(farm, buildNumber, branch, commit string) Option {
 	return func(c *config) {
-		c.farm = farm
-		c.buildNumber = buildNumber
-		c.branch = branch
-		c.commit = commit
+		c.tag("farm", farm)
```
jamestelfer, line 96: Build details are now just added to the `tags` map on the config struct, rather than being treated specially.

melm73: Nice!!

melm73 (approved):

jamestelfer:
Thanks @melm73!

PR #32, opened 2022-06-06
Basic Sarama client configuration for Kafka

jamestelfer:
Provides a simple way to sets up the Sarama Kafka client for Go such that it uses SASL with SHA512 and TLS.


aimeegee (approved):
Thanks JT!

PR #33, opened 2022-06-07
COSE-90 Add support for targeting surveys in LaunchDarkly

El-Wel:
## Purpose
To get more flexibility out of the platform, our implementation should support evaluation contexts that map to entities other than human users. Some of our use cases revolve around the surveys domain, which can be its own evaluation context and be treated as a first-class citizen of the platform. This means we can target flags against segments of surveys.

## Context
In addition to adding survey evaluation context, this also introduces submission of userID and surveyID custom attributes with user and survey contexts respectively in order to provide an easy way of identifying type of context in LaunchDarkly UI.

This PR is consistent with the [Node](https://github.com/cultureamp/ca-launchdarkly-js/pull/68) and [Ruby](https://github.com/cultureamp/ca-launchdarkly-ruby/pull/33) Wrappers
[JIRA Ticket](https://cultureamp.atlassian.net/browse/COSE-90?atlOrigin=eyJpIjoiYThmZjY0OGZkNjQ0NGViYWFhOTZiN2M3NDVjMWQ3MTEiLCJwIjoiaiJ9)

### DIFF
--- a/x/launchdarkly/flags/evaluationcontext/doc.go
+++ b/x/launchdarkly/flags/evaluationcontext/doc.go
@@ -17,9 +17,4 @@
 // attributes as you can to give yourself more flexibility when writing new
 // targeting rules. When you query a flag containing a rule that works on
 // attribute "foo", you must supply attribute "foo" in the evaluation context.
-//
-// The constructor functions will namespace the key of the evaluation context
-// you send in the query. You do not need to prefix the values provided to
-// the constructor functions with the entity type. For example, supply the user
-// ID as-is rather than prefixing as `user.<user-id>`.
 package evaluationcontext
--- a/x/launchdarkly/flags/evaluationcontext/survey.go
+++ b/x/launchdarkly/flags/evaluationcontext/survey.go
@@ -1,0 +1,41 @@
+package evaluationcontext
+
+import (
+	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+)
+
+const (
+	surveyAttributeSurveyID = "surveyID"
+)
+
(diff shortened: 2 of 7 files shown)


x/launchdarkly/flags/evaluationcontext/doc.go
```diff
@@ -18,8 +18,7 @@
 // targeting rules. When you query a flag containing a rule that works on
 // attribute "foo", you must supply attribute "foo" in the evaluation context.
 //
-// The constructor functions will namespace the key of the evaluation context
-// you send in the query. You do not need to prefix the values provided to
-// the constructor functions with the entity type. For example, supply the user
-// ID as-is rather than prefixing as `user.<user-id>`.
+// You do not need to prefix the values provided to the constructor functions
+// with the entity type. For example, supply the user ID as-is rather than
+// prefixing as `user.<user-id>`.
```
hallaji, line 23: This doc's a bit confusing. Because prefixing is not a thing anymore, shall we remove it entirely?

El-Wel: Yep, makes sense! Have updated :)

x/launchdarkly/flags/evaluationcontext/survey.go
```diff
@@ -0,0 +1,41 @@
...
+	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+)
+
+const (
+	surveyAttributeSurveyID = "surveyID"
+)
+
+// Survey is a type of context, representing the identifiers and attributes of
+// a human user to evaluate a flag against.
```
einal3m, line 13: Should this be "survey" rather than "human user"?

einal3m (approved):

PR #34, opened 2022-06-24
add more sarama kafka config options to help testing

aimeegee:
## Purpose
Currently we only provide a default config which won't work with local Kafka. 

## Context
- Added a function to turn on sarama debug log to provide more details when connecting brokers failed
- Added a non-SSL config function to return sarama config that could be used for local Kafka


einal3m (approved):
Nice!

disposedtrolley:
:shipit:

PR #36, opened 2022-07-21
[COSE-319] Add secret manager fetching to ca-go.

einal3m:
This PR extracts our secret fetching code out of the `accounts-service` and `fusionauth-hooks`. We are about to require this logic for the 3rd time, so thought it would be good to extract it out here.

This new package creates the secret manager client for a given region, and fetches the secret value given a secret name.


hughevans:
I think we should also adding an `x/secrets/README.md` file that details how you should use this to only pull the secrets on cold start.

einal3m:
> I think we should also be adding a `x/secrets/README.md` file that details how you should use this to only pull the secrets on cold start.

I've added more information to the `doc.go`, let me know if you think that's enough

hughevans (approved):

PR #38, opened 2022-07-28
feat: configure big segments and use refined config structure

hallaji:
The wrapper is now configured to get Big Segments data from the Relay Proxy persistent storage regardless of the connection mode. In order for us to actually support big segments, we need to connect our clients directly to the data table, and that requires DynamoDB `GetItem` permission on relay table.

### DIFF
--- a/x/launchdarkly/flags/config.go
+++ b/x/launchdarkly/flags/config.go
@@ -27,14 +27,12 @@ const (
 // environment variable.
 type configurationJSON struct {
 	SDKKey  string `json:"sdkKey"`
-	Options struct {
-		DaemonMode *struct {
-			DynamoTableName string `json:"DynamoTableName"`
-		} `json:"daemonMode"`
-		Proxy *struct {
-			RelayProxyURL string `json:"url"`
-		} `json:"proxyMode"`
-	} `json:"options"`
+	Storage *struct {
+		TableName string `json:"tableName"`
+	} `json:"storage"`
+	Proxy *struct {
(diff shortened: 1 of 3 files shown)


x/launchdarkly/flags/config.go
```diff
@@ -27,14 +27,12 @@ const (
...
 	SDKKey  string `json:"sdkKey"`
-	Options struct {
-		DaemonMode *struct {
-			DynamoTableName string `json:"DynamoTableName"`
-		} `json:"daemonMode"`
-		Proxy *struct {
-			RelayProxyURL string `json:"url"`
-		} `json:"proxyMode"`
-	} `json:"options"`
+	Storage *struct {
```
aimeegee, line 30: I might not have enough background knowledge this. But this looks like a broken change so is it possible to receive old config json? Do we need to consider backward compatibility here?

hallaji: That's a good point, Aimee! I'm not quite sure how versioning works in this repo (commit SHA?) 
The secrets structure has been already updated (those options are still in the secret, and it's still backward compatible). We are introducing a new version in source code, and I guess previously used versions won't be touched. Is the assumption correct regarding versioning?

Allen-Ning: **The secrets structure has been already updated (those options are still in the secret, and it's still backward compatible)**

Hi @hallaji, I think this will be backward compatible.

Allen-Ning (approved):
nice work!!!

PR #41, opened 2022-08-10
[core-services] added new anonymous user with subdomain function for web gateway

Allen-Ning:
Added new anonymous user with subdomain function for web-gateway

Similar to `NewAnonymousUser` function, we would like to capture user subdomain as one custom attribute so that devs can toggle the feature toggle to filter out anonymous user within a particular subdomain (e.g. `qastandard`)

### DIFF
--- a/x/launchdarkly/flags/evaluationcontext/user.go
+++ b/x/launchdarkly/flags/evaluationcontext/user.go
@@ -14,6 +14,7 @@ const (
 	userAttributeUserID     = "userID"
 	userAttributeAccountID  = "accountID"
 	userAttributeRealUserID = "realUserID"
+	userAttributeSubdomain  = "subdomain"
 )
 
 // User is a type of context, representing the identifiers and attributes of
@@ -62,15 +63,34 @@ func NewAnonymousUser(key string) User {
 		key = uuid.NewString()
 	}
 
-	u := User{
+	return User{
 		key: key,
+		ldUser: lduser.NewUserBuilder(key).
+			Anonymous(true).
+			Build(),
+	}
+}
+
+// NewAnonymousUserWithSubdomain returns a user object suitable for use in unauthenticated with known subdomain
+// requests or requests with no access to user identifiers.
+// Provide a unique session or request identifier as the key if possible. If the
+// key is empty, it will default to an uuid so percentage rollouts will still apply.
+// No userID will be given to an anonymous user.
+func NewAnonymousUserWithSubdomain(key string, subdomain string) User {
+	if key == "" {
+		key = uuid.NewString()
 	}
 
-	userBuilder := lduser.NewUserBuilder(u.key)
-	userBuilder.Anonymous(true)
-	u.ldUser = userBuilder.Build()
-
-	return u
+	return User{
+		key: key,
+		ldUser: lduser.NewUserBuilder(key).
+			Anonymous(true).
+			Custom(
+				userAttributeSubdomain,
+				ldvalue.String(subdomain),
+			).
+			Build(),
+	}
 }
 
 // NewUser returns a new user object with the given user ID and options.
--- a/x/launchdarkly/flags/evaluationcontext/user_test.go
+++ b/x/launchdarkly/flags/evaluationcontext/user_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/cultureamp/ca-go/x/request"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
 )
 
 func TestNewUser(t *testing.T) {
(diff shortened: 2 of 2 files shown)


x/launchdarkly/flags/evaluationcontext/user.go
```diff
@@ -73,6 +74,30 @@ func NewAnonymousUser(key string) User {
...
+	}
+
+	userBuilder := lduser.NewUserBuilder(u.key)
+	userBuilder.Anonymous(true)
+	userBuilder.Custom(
+		userAttributeSubdomain,
+		ldvalue.String(subdomain))
+	u.ldUser = userBuilder.Build()
+
+	return u
```
joshjon, line 98: This can be simplified.

```suggestion
	return User{
		key: key,
		ldUser: lduser.NewUserBuilder(key).
			Anonymous(true).
			Build(),
	}
```

x/launchdarkly/flags/evaluationcontext/user_test.go
```diff
@@ -22,6 +23,26 @@ func TestNewUser(t *testing.T) {
...
+		user := evaluationcontext.NewAnonymousUserWithSubdomain("", "cultureamp")
+		ldUser := user.ToLDUser()
+		assert.True(t, ldUser.GetAnonymous())
+
+		value, result := ldUser.GetCustom("subdomain")
+		assert.Equal(t, ldvalue.String("cultureamp"), value)
+		assert.True(t, result)
+	})
+
+	t.Run("can create an anonymous user with subdomain", func(t *testing.T) {
```
hallaji, line 36: Test description seems to be same as above.

x/launchdarkly/flags/evaluationcontext/user.go
```diff
@@ -62,15 +63,34 @@ func NewAnonymousUser(key string) User {
...
 
-	userBuilder := lduser.NewUserBuilder(u.key)
-	userBuilder.Anonymous(true)
-	u.ldUser = userBuilder.Build()
+// NewAnonymousUserWithSubdomain returns a user object suitable for use in unauthenticated with known subdomain
+// requests or requests with no access to user identifiers.
+// Provide a unique session or request identifier as the key if possible. If the
+// key is empty, it will default to an uuid so percentage rollouts will still apply.
+// No userID will be given to an anonymous user.
+func NewAnonymousUserWithSubdomain(key string, subdomain string) User {
```
joshjon, line 79: One thing to consider about adding this constructor is that it beings to break the Functional Options pattern used above via`type UserOption func(*User)`. I wanted to make a suggestion in a seperate branch to keep this PR clean, but I don't have editor permissions to push it up. I've copied the main chunk of it below for reference instead. Basically, it looks like things can be restructured a bit to support anonymous user and subdomain as options.

```go
// User is a type of context, representing the identifiers and attributes of
// a human user to evaluate a flag against.
type User struct {
	anonymous bool
	builder   lduser.UserBuilder
}

func (u User) ToLDUser() lduser.User {
	return u.builder.Build()
}

// UserOption are functions that can be supplied to configure a new user with
// additional attributes.
type UserOption func(*User)

// WithAccountID configures the user with the given account ID.
// This is the ID of the currently logged in user's parent account/organization,
// sometimes known as the "account_aggregate_id".
func WithAccountID(id string) UserOption {
	return func(u *User) {
		u.builder.Custom(userAttributeAccountID, ldvalue.String(id))
	}
}

// WithRealUserID configures the user with the given real user ID.
// This is the ID of the user who is currently impersonating the current user.
func WithRealUserID(id string) UserOption {
	return func(u *User) {
		u.builder.Custom(userAttributeRealUserID, ldvalue.String(id))
	}
}

func WithAnonymous() UserOption {
	return func(u *User) {
		u.builder.Anonymous(true)
	}
}

func WithSubDomain(subdomain string) UserOption {
	return func(u *User) {
		u.builder.Custom(userAttributeSubdomain, ldvalue.String(subdomain))
	}
}

// NewUser returns a new user object with the given user ID and options.
// userID is the ID of the currently authenticated user, and will generally
// be a "user_aggregate_id".
func NewUser(userID string, opts ...UserOption) User {
	u := User{
		builder: lduser.NewUserBuilder(userID),
	}

	for _, opt := range opts {
		opt(&u)
	}

	if !u.anonymous {
		u.builder.Custom(
			userAttributeUserID,
			ldvalue.String(userID))
	}

	return u
}
```

Ultimately, I personally don't mind if you keep it the way it is 🙂 but the usefulness of the functional options pattern might be lost a bit.

Allen-Ning: Good observation. I was thinking about `UserOption pattern` as well. But I think for us, `domain` is currently only one custom user case for `web-gateway`. But the rest of them (`AccountID`, `UserID`, `RealUserID`)  are pretty much can be found in any backend services given they are value in our JWT token.

joshjon: Fair enough, if it's only for a single use case then we can leave it as is 👍

joshjon (approved):
LGTM

hallaji (approved):

PR #42, opened 2022-08-15
[COSE-418] Add vault decryption module

El-Wel:
## Purpose
As decryption using vault is likely to be implemented in many areas, it is being extracted into ca-go to centralise and standardise it. 

## Context
Unit testing is done on client and en/decryption with the intention that integration testing will be performed on the consumer end. Integration testing will be done with an external service locally before merging the PR.


x/launchdarkly/flags/client.go
```diff
@@ -91,7 +91,7 @@ func NewClient(opts ...ConfigOption) (*Client, error) {
 // connection error occurs.
 func (c *Client) Connect() error {
 	if c.wrappedClient != nil {
-		return errors.New("attempted to call Connect on a connected client")
```
einal3m, line 94: This is the launchdarkly code... I don't think this is right?! 🤔

x/vaultdecryption/doc.go
```diff
@@ -0,0 +1,12 @@
...
+// settings := client.VaultSettings{
+////				DecrypterRoleArn: <arn here>,
+////				VaultAddr:        <vault address here>,
+////			}
+// client, err :=  client.NewVaultClient(&settings, ctx context.Context)
+//
+// decrypter := NewVaultDecrypter(client, &settings})
+//
+// decryptedSecret, err := decrypter.Decrypt(keyReferences, encryptedData, ctx)
+package vaultdecryption
```
einal3m, line 12: I wonder if the package could be called just `vault` and it can provide both encryption and decryption (eventually). This is under the assumption that there will be quite a bit of common code. If not ignore me.

x/vaultdecryption/client/vault_client.go
```diff
@@ -0,0 +1,87 @@
...
+	DecrypterRoleArn string
+	VaultAddr        string
+}
+
+type VaultClient struct {
+	settings *VaultSettings
+	client   *vaultapi.Client
+}
+
+func NewVaultClient(settings *VaultSettings, ctx context.Context) (*VaultClient, error) {
```
joshjon, line 33: *VaultSettings probably doesn't need to be a pointer since its just static config.

x/vaultdecryption/client/vault_client.go
```diff
@@ -0,0 +1,87 @@
...
+
+const (
+	vaultDecrypterRole   = "decrypter"
+	VaultPermissionError = "Code: 403"
+)
+
+var (
+	Create    = NewVaultClient
+	Login     = wrappedLogin
+	NewClient = wrappedNewVClient
```
joshjon, line 20: Whats the purpose of these vars being assigned to functions? Is this purely to override for tests?

x/vaultdecryption/vault_decrypter.go
```diff
@@ -0,0 +1,121 @@
...
+type vaultDecrypter struct {
+	vaultClient Client
+	settings    *client.VaultSettings
+}
+
+func NewVaultDecrypter(vaultClient Client, settings *client.VaultSettings) *vaultDecrypter {
+	return &vaultDecrypter{vaultClient, settings}
+}
+
+func (v vaultDecrypter) Decrypt(keyReferences []string, encryptedData []string, ctx context.Context) ([]string, error) {
```
joshjon, line 32: `NewVaultDecrypter` returns a pointer of vaultDecrypter, so the methods on the struct should probably take in a pointer receiver.

```suggestion
func (v *vaultDecrypter) Decrypt(keyReferences []string, encryptedData []string, ctx context.Context) ([]string, error) {
```

x/vaultdecryption/vault_decrypter.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/cultureamp/glamplify/log"
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+type Client interface {
+	RenewClient(ctx context.Context) error
+	GetSecret(batch []interface{}, keyReference string) (*vaultapi.Secret, error)
+}
+
+type VaultDecrypter interface {
```
joshjon, line 19: I can't seem to find where this Interface is referenced. Is this added purely so that people using the package can utilise this interface? If that's the case it can probably be removed since it can be the responsibility of the caller to add an interface if they require it. You can also export `vaultDecrypter` if you'd like after that.

joshjon (commented):
Nice clean code 👌, just left some minor comments.

x/vaultdecryption/auth/aws_iam_auth.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-secure-stdlib/awsutil"
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+const (
+	defaultStsRegion = "us-east-1"
+	loginPath        = "auth/aws/login"
+)
```
Allen-Ning, line 17: i wonder if we could load these two values loaded from Env, and use "us-east-1", "auth/aws/login" as a default value.

x/vaultdecryption/auth/aws_iam_auth.go
```diff
@@ -0,0 +1,56 @@
...
+		roleArn:   roleArn,
+	}
+}
+
+func (a *AWSIamAuth) Login(ctx context.Context, client *vaultapi.Client) (*vaultapi.Secret, error) {
+	var awsSession *session.Session
+	var err error
+	if awsSession, err = session.NewSession(); err != nil {
+		return nil, fmt.Errorf("failed to create AWS session: %w", err)
+	}
```
Allen-Ning, line 36: how about removing the `var awsSession *session.Session`,  `var err error`
awsSession, err := session.NewSession(); err != nil {
  return nil, fmt.Errorf("failed to create AWS session: %w", err)
}

x/vaultdecryption/auth/aws_iam_auth.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func NewAWSIamAuth(vaultRole string, roleArn string) *AWSIamAuth {
+	return &AWSIamAuth{
+		vaultRole: vaultRole,
+		roleArn:   roleArn,
+	}
+}
+
+func (a *AWSIamAuth) Login(ctx context.Context, client *vaultapi.Client) (*vaultapi.Secret, error) {
```
Allen-Ning, line 31: how about use `auth` replace `a` to be more meaningful?

Allen-Ning (commented):
looks good. One thing we can improve here is to mock those AWS functions by adding your own small wrapper with interface, which will make you way more easy to mock without overwriting functions.

x/vaultdecryption/auth/aws_iam_auth.go
Allen-Ning, line 31: looks like we might miss a test to cover functions in this file

x/vaultdecryption/vault_decrypter.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return nil, err
+		}
+		result = decryptedByKeyReference
+	}
+
+	return result, nil
+}
+
+func (v vaultDecrypter) decryptByKey(keyReference string, encryptedData []string, logger log.Logger, ctx context.Context) ([]string, error) {
```
Allen-Ning, line 47: How about just pass reference for `logger` here?

x/vaultdecryption/vault_decrypter.go
```diff
@@ -0,0 +1,121 @@
...
+
+func NewVaultDecrypter(vaultClient Client, settings *client.VaultSettings) *vaultDecrypter {
+	return &vaultDecrypter{vaultClient, settings}
+}
+
+func (v vaultDecrypter) Decrypt(keyReferences []string, encryptedData []string, ctx context.Context) ([]string, error) {
+	logger := log.NewFromCtx(ctx)
+
+	result := encryptedData
+	for _, keyReference := range reverse(keyReferences) {
```
Allen-Ning, line 36: feeling like no need to do a real reverse. I guess you can directly do a reverse loop here. Please feel free to ignore this.

x/vaultdecryption/vault_decrypter.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (v vaultDecrypter) decryptWithVault(keyReference string, batch []interface{}, logger log.Logger, ctx context.Context) (*vaultapi.Secret, error) {
+	var secret *vaultapi.Secret
+	var err error
+	maxRetries := 5
+	for i := 0; i < maxRetries; i++ {
+		secret, err = v.vaultClient.GetSecret(batch, keyReference)
+		if err != nil {
+			if strings.Contains(err.Error(), client.VaultPermissionError) {
```
Allen-Ning, line 95: just curiosity here, why we retry 5 times only for permission errors? feel like most of time there is no need to retry permission error

x/launchdarkly/flags/client.go
El-Wel: Oh dear, I would say a rogue refactor has gone awry 😳
Good catch, I will fix that up!

El-Wel:
Feedback was provided on Slack to also use integration testing to ensure future changes don't affect the behaviour of the code. In addition, it was suggested to introduce datadog tracing by receiving the span from upstream services.

x/vaultdecryption/auth/aws_iam_auth.go
El-Wel: I think the `loginPath` remains the same but I think having the sts region set though env var might be worthwhile

x/vaultdecryption/client/vault_client.go
El-Wel: Yeah this was for overriding the functions during testing

El-Wel:
Differentiate between decryption error and other error types for deleted keys

go.mod
```diff
@@ -12,34 +12,91 @@ require (
 
 require (
 	github.com/Shopify/sarama v1.34.0
+	github.com/cultureamp/glamplify v1.6.0
```
hughevans, line 15: We can switch this to Zerolog now, see #52

El-Wel: I have added the module and it is very cool. Love the colour coded messages in local!

x/vault/client/vault_client.go
```diff
@@ -0,0 +1,140 @@
...
+	VaultPermissionError = "Code: 403"
+	EncryptionAction     = "encrypt"
+	DecryptionAction     = "decrypt"
+)
+
+var (
+	Create                = NewVaultClient
+	Login                 = wrappedLogin
+	NewClient             = wrappedNewVClient
+	VaultMissingKeysError = errors.New("no key references passed")
```
aimeegee, line 30: Let's try rename this to `VaultMissingKeysErr`

einal3m (approved):

PR #55, opened 2022-08-29
[core-services] fix LD initialisation issue

Allen-Ning:
We're noticing this is an issue when multiple thread or go-routines accessing the LD client at the same time. 

`runtime error: invalid memory address or nil pointer dereference`

The issue is introduced by LD client (initialising in progress) not ready for serving requests. 

A few improvements are listed below: 
- [x] Remove Connect, which means only ready-to-serve client is exposed
- [x] Options to disable big segments

### Context
ticket: https://cultureamp.atlassian.net/browse/COSE-476

### DIFF
--- a/x/launchdarkly/flags/client.go
+++ b/x/launchdarkly/flags/client.go
@@ -14,11 +14,12 @@ import (
 
 // Client is a wrapper around the LaunchDarkly client.
 type Client struct {
-	sdkKey        string
-	initWait      time.Duration
-	mode          mode
-	wrappedConfig ld.Config
-	wrappedClient *ld.LDClient
+	sdkKey             string
+	initWait           time.Duration
+	mode               mode
+	bigSegmentsEnabled bool
+	wrappedConfig      ld.Config
+	wrappedClient      *ld.LDClient
 
 	testModeConfig *TestModeConfig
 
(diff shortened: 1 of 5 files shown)


x/launchdarkly/flags/client.go
```diff
@@ -42,17 +43,25 @@ const (
 // launchdarkly/flags/doc.go for more information.
 func NewClient(opts ...ConfigOption) (*Client, error) {
 	c := &Client{
-		initWait: 5 * time.Second, // wait up to 5 seconds for LD to connect.
-		mode:     modeProxy,       // defaults to proxying requests through the LD Relay.
+		initWait:           5 * time.Second, // wait up to 5 seconds for LD to connect.
+		mode:               modeProxy,       // defaults to proxying requests through the LD Relay.
+		bigSegmentsEnabled: false,           // defaults to disable big segments
```
Allen-Ning, line 48: disable by default. Let me know if think you should get it enabled by default

x/launchdarkly/flags/client.go
```diff
@@ -42,17 +43,25 @@ const (
...
 	}
 
 	for _, opt := range opts {
 		opt(c)
 	}
 
-	// Use test mode if LAUNCHDARKLY_CONFIGURATION isn't set OR if the user
-	// explicitly configured the client for test mode.
-	if _, ok := os.LookupEnv(configurationEnvVar); !ok || c.mode == modeTest {
+	parsedConfig := configurationJSON{}
```
Allen-Ning, line 55: I would like to clearly seperate the logic for `parsedConfig` and `modeTest` for easy testing purpose. 
For now,  

1.  `configurationEnvVar` will be checked as long as it is set
2. test mode needs to be clearly and explicitly configured the client for test mode

x/launchdarkly/flags/flags.go
```diff
@@ -16,19 +16,12 @@ func Configure(opts ...ConfigOption) error {
...
-// been established, or a connection error occurs.
-func Connect() error {
-	if flagsClient == nil {
-		return errClientNotConfigured
+	if err := c.Connect(); err != nil {
+		return fmt.Errorf("connect client: %w", err)
 	}
 
-	return flagsClient.Connect()
+	flagsClient = c
```
Allen-Ning, line 23: we only would like to set `flagsClient = c` when c is ready to reserve request

x/launchdarkly/flags/client.go
hallaji: Generally speaking, `false` default makes sense to me. However, users are intended to use big segments for most of the cases, so it's enabled in the other wrappers by default. Are you planning to disable it in your web-gateway use case?

Allen-Ning: > Generally speaking, `false` default makes sense to me. However, users are intended to use big segments for most of the cases, so it's enabled in the other wrappers by default. Are you planning to disable it in your web-gateway use case?
Yep. I'm planning to disable it in your web-gateway

hallaji (approved):
LGTM 👋🏼

x/launchdarkly/flags/flags.go
hallaji: Shall we set this to true by default?

Allen-Ning: > Shall we set this to true by default?

is nil by default?

hallaji: Sorry wrong place. I meant to post on big segment one. 😄

x/launchdarkly/flags/client.go
hallaji: I meant to post it here, Shall we set this to true by default?

Allen-Ning: > users are intended to use big segments for most of the cases, so it's enabled in the other wrappers by default

Yes. We can.

hughevans: Thanks, yes I agree that default `true` is the way to go 👍

Allen-Ning: updated

hallaji (approved):

PR #58, opened 2022-09-05
[core-services] cose-476 generalise ld golang wrapper for common fields

Allen-Ning:


x/launchdarkly/flags/evaluationcontext/survey.go
```diff
@@ -22,19 +23,35 @@ func (u Survey) ToLDUser() lduser.User {
 	return u.ldUser
 }
 
+func (u *Survey) SetAccountID(accountID string) {
+	u.accountID = accountID
+}
+
+func (u *Survey) SetRealUserID(realUserID string) {
```
Allen-Ning, line 30: otherwise, we can split this into two different options to avoid this 
func NewUser(userID string, sharedOpts []Option, userOpts []UserOption) User
func NewSurvey(userID string, sharedOpts []Option, surveyOpts []SurveyOption) Survey

PR #67, opened 2022-09-13
[core-services] CORE 526 - add gracefully shutdown functionality to ca go for sentry

Allen-Ning:
If the our service is panic or fatal. We need to send those causing failure events to sentry server as well.

- [x] added gracefullyShutdown function

### Context
ticket - https://cultureamp.atlassian.net/browse/COSE-526

### DIFF
--- a/x/sentry/errorreport/errorreport.go
+++ b/x/sentry/errorreport/errorreport.go
@@ -2,8 +2,10 @@ package errorreport
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/cultureamp/ca-go/x/request"
 	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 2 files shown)


hughevans (approved):
Nice one. Should we include how to use it in the docs, or make another card/PR for that?

hughevans (dismissed):

x/sentry/errorreport/errorreport.go
```diff
@@ -93,6 +95,22 @@ func Decorate(tags map[string]string) func() {
...
+	switch content := err.(type) {
+	case string:
+		newError = errors.New(content)
+	case error:
+		newError = content
+	default:
+		message := fmt.Sprintf("unknown panic: %T", content)
+		newError = errors.New(message)
+	}
+	sentry.CurrentHub().Recover(newError)
```
Allen-Ning, line 110: by testing, it seem like sentry can only recover from error. It err is a string type, we need to cover it to error type.

hughevans (approved):

PR #74, opened 2022-09-19
add a small doc to show how to use sentry in Fargate service in ca go

Allen-Ning:
add a small doc to show how to use sentry in Fargate service in ca go

### Context
Ticket: https://cultureamp.atlassian.net/browse/COSE-527

### DIFF
--- a/x/sentry/errorreport/examples_test.go
+++ b/x/sentry/errorreport/examples_test.go
@@ -3,13 +3,14 @@ package errorreport_test
 import (
 	"context"
 	"os"
+	"time"
 
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/cultureamp/ca-go/x/log"
 	"github.com/cultureamp/ca-go/x/sentry/errorreport"
 )
 
-// Set these variables at build time using linker flags (-ldflags)
 var (
 	app         string
 	appVersion  string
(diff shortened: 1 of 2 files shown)


joshjon (approved):
LGTM

El-Wel:
The documentation looks good but using a readme file isn’t consistent with the existing documentation in this repo, might I suggest adding this to the existing documentation in this directory?

x/sentry/errorreport/examples_test.go
```diff
@@ -94,3 +99,55 @@ func processRecord(ctx context.Context, record events.KinesisEventRecord) error
 	// do something with the record
 	return nil
 }
+
+func Example_second() {
+	// This is an example of how to use the errorreport package in a Main
```
El-Wel, line 104: The above example is also of a main function but specifies it is for a lambda, is this example of a fargate main?

x/sentry/errorreport/examples_test.go
```diff
@@ -94,3 +99,55 @@ func processRecord(ctx context.Context, record events.KinesisEventRecord) error
 	// do something with the record
 	return nil
 }
+
+func Example_second() {
```
El-Wel, line 103: The name of this function will show on go doc as Example (Second) (attached screenshot), could these two examples be named to describe what they are doing like Example (Lambda) and Example (Fargate)? Or are they both just examples of the same thing?
Screen Shot 2022-09-19 at 2 32 14 pm

Allen-Ning: > The name of this function will show on go doc as Example (Second) (attached screenshot), could these two examples be named to describe what they are doing like Example (Lambda) and Example (Fargate)? Or are they both just examples of the same thing? Screen Shot 2022-09-19 at 2 32 14 pm

Unfortunately, you can only name them `example`, `example_second`, `example_third` for package level examples.
E.g. `example_for_lambda` will not work

El-Wel: Ah dang, I was able to use it locally. Is this feature just not supported on pkg.go.dev?
Screen Shot 2022-09-19 at 2 40 20 pm

El-Wel: It seems that `Example_for_fargate` does not work but `Example_fargate` does. Looks like it just can't support more than one additional word

Allen-Ning: > It seems that `Example_for_fargate` does not work but `Example_fargate` does. Looks like it just can't support more than one descriptive word

Yes. The naming convention is very limited. Happy to change it to `Example_xxx` if it makes more sense for u.
Screen Shot 2022-09-19 at 2 48 16 pm

El-Wel: I think having a descriptive suffix is more helpful, especially as this is a public repository

El-Wel (approved):
Thanks for the discussion and updates! Looks great!

PR #83, opened 2022-11-28
[COSE-470] Feature: add Kafka consumer package

joshjon:
This PR adds a Kafka consumer package that is built on top of [segmentio/kafka-go](https://github.com/segmentio/kafka-go).

The package contains two main types:

- `Consumer`: provides a high level API for consuming and handling messages from a Kafka topic.
- `Group`: a group of consumers used to concurrently consume and handle messages from a Kafka topic.

This package has been successfully used in prod by team Core Services as part of the FusionAuth Employee Consumer.


x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,321 @@
...
+// Close closes the consumer, preventing it from consuming any more messages.
+func (c *Consumer) Close() error {
+	defer func() { c.closed = true }()
+	if err := c.reader.Close(); err != nil {
+		return fmt.Errorf("unable to close consumer %s: %w", c.id, err)
+	}
+	return nil
+}
+
+func (c *Consumer) consumeMessage(ctx context.Context, handler Handler) error {
```
Allen-Ning, line 120: It seems like there are two different patterns for committing a message - auto commit vs explicit commit 

```
m, err := r.ReadMessage(context.Background())
```

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,321 @@
...
+
+// Metadata contains relevant handler metadata for received Kafka messages.
+type Metadata struct {
+	GroupID    string
+	ConsumerID string
+	Attempt    int
+}
+
+// NotifyError is a notify-on-error function used to report consumer handler errors.
+type NotifyError func(ctx context.Context, err error, msg kafka.Message, metadata Metadata)
```
Allen-Ning, line 29: how about build a custom Message struct?

```
type Message {
  msg kafka.Message
  metadata Metadata
}

type NotifyError func(ctx context.Context, err error,  msg Message)
```

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	c := &Consumer{
+		id: config.ID,
+		readerConfig: kafka.ReaderConfig{
+			Brokers:               config.Brokers,
+			GroupID:               config.GroupID,
+			Topic:                 config.Topic,
+			Dialer:                dialer,
+			GroupBalancers:        []kafka.GroupBalancer{kafka.RoundRobinGroupBalancer{}},
```
Allen-Ning, line 74: Given this is the template for everyone in CA, it seems like it is worth to add the group balancers in into config as well.

x/kafka/consumer/consumer.go
joshjon: I like this - will make the change

x/kafka/consumer/consumer.go
joshjon: Good call out, will do

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+
+		attempt++
+
+		consumerMsg := Message{
+			Message: msg,
+			Metadata: Metadata{
+				GroupID:    c.readerConfig.GroupID,
+				ConsumerID: c.id,
+				Attempt:    attempt,
```
Allen-Ning, line 187:
```suggestion
				Attempt:    attempt + 1,
```

x/kafka/consumer/consumer.go
joshjon: Implemented in revision https://github.com/cultureamp/ca-go/pull/83/commits/7b625ec6bf20f92000b6e4b39a58ebe2f373ef98

x/kafka/consumer/consumer.go
joshjon: Implemented in revision https://github.com/cultureamp/ca-go/commit/7b625ec6bf20f92000b6e4b39a58ebe2f373ef98

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,329 @@
...
+		} else {
+			c.reader = kafka.NewReader(c.readerConfig)
+		}
+	}
+
+	return c
+}
+
+// Run consumes and handles messages from the topic. The method call blocks until
+// the context is canceled, the consumer's reader is closed, or an error occurs.
```
Allen-Ning, line 100: just curious here `the context is canceled`, how do we trigger the `cancel` if needed?

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,329 @@
...
+			return fmt.Errorf("unable to extract data dog span context from kafka message: %w", err)
+		}
+		span := tracer.StartSpan("consumer.handle", tracer.ChildOf(spanCtx))
+		defer span.Finish()
+		ctx = tracer.ContextWithSpan(ctx, span)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
```
Allen-Ning, line 169: My understanding is that: If the given context is `canceled` or `timeout`, the `<-ctx.done()` will be executed. But 
similar to my other question, I cannot find any code could potential `trigger the context cancel or timeout`

x/kafka/consumer/consumer.go
joshjon: Here is an example: 

```go
ctx, cancel := context.WithCancel(context.Background())
go func() {
    time.Sleep(10*time.Second)
    cancel()
}()
c := consumer.NewConsumer(kafka.DefaultDialer, groupCfg)
defer c.Close()
if err := c.Run(ctx, handle); err != nil {
    panic(err)
}
```

`c.Run` would be stopped after 10s due to the cancel fn being invoked

x/kafka/consumer/consumer.go
joshjon: Implemented in revision https://github.com/cultureamp/ca-go/pull/83/commits/70b6887b8d02fa2b937e49c14427734d4c92a9da

x/kafka/consumer/consumer.go
Allen-Ning: it would be great to add this example in `example/consumer/main.go`

x/kafka/consumer/consumer.go
Allen-Ning: Feel free to ignore this given it has been answered in another thread

x/kafka/consumer/consumer.go
amatueva: Maybe worth adding handling of `SIGINT/SIGTERM` in the example too?

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,329 @@
...
+			if err == nil {
+				return ctx.Err()
```
amatueva, line 179: I don't understand this 100%, will this result in error being returned? I saw in the example that if `consumer.Run()` returns error it's going to panic? What if this a legitimate case when consumer needed to be shutdown

joshjon: We return the error here because we explicitly want to let the caller know that we have stopped due to a cancelled context. If we were to instead return nil then it would be very misleading behaviour.

With that said, it is 100% up to the caller to check the returned error and to handle accordingly e.g.

```go
if err := c.Run(context.Background(), handle); err != nil {
	if err != nil {
		if errors.Is(err, context.Canceled) {
			log.Println("context cancelled")
		} else if errors.Is(err, context.DeadlineExceeded) {
			log.Println("deadline exceeded")
		}
		panic(err)
	}
}
```

x/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,338 @@
...
+				}
+				return fmt.Errorf("consumer %s unable to read message: %w", c.id, err)
+			}
+		}
+
+		if err = c.handle(ctx, msg, handler); err != nil {
+			return fmt.Errorf("consumer %s unable to handle message: %w", c.id, err)
+		}
+
+		if c.withExplicitCommit {
```
amatueva, line 135: Should explicit commit be in the handle function perhaps? If user were to pick to drop the corrupted message/publish to dead letter queue they would have an ability to do so if they don't want infinite retry?

x/kafka/consumer/consumer.go
joshjon: I added in a SIG channel to the consumer group example, but I think it's better to leave out context cancelling in order to avoid overcomplicating. 

Context cancelling is a pretty foundational features in Go and I don't think its necessarily the responsibility of the package to show how it works in practice, especially when there's a variety of ways to cancel a context using `context.WithCancel`, `context.WithTimeout`, `context.WithDeadline`.

For instance, refer to the [kafka-go ](https://github.com/segmentio/kafka-go/blob/main/examples/consumer-logger/main.go) example and you will notice its extremely simple. The kafka-go reader supports context cancelling as well, yet it doesn't show how it works in the example. Simply acknowledging the support for it in the docs is enough in my opinion.

x/kafka/consumer/consumer.go
amatueva: I feel like it could be a bit misleading that every redeploy will result with an error when consumer actually successfully shutdown but don't have a strong opinion on this

x/kafka/consumer/consumer.go
joshjon: Sorry I don't think I understand your comment, would you mind elaborating a bit more?

Publishing to a dead letter queue is supported with the default behaviour of auto committing offsets on read

x/kafka/consumer/consumer.go
joshjon: > I feel like it could be a bit misleading that every redeploy will result with an error when consumer actually successfully shutdown but don't have a strong opinion on this

I'm not sure how this relates to returning `return ctx.Err()`?

x/kafka/consumer/consumer.go
amatueva: If handler errors on line 131 it will never reach line 135 so it will infinitely retry till user corrects handler function, right? I mean it's a choice too, I guess

x/kafka/consumer/consumer.go
amatueva: Would you call context cancel on redeploy (if you handle SIGINT and SIGTERM correctly) to notify other processes to return?

x/kafka/consumer/consumer.go
joshjon: I guess the way I look at it is that the user would implement the logic to publish to a dead letter queue within the handler and return no error. Otherwise we would need to implement all the logic to do this for them, which I was hoping to avoid.

The handler also wouldn't infinitely retry, it would just return an error and stop processing (unless configured with a retry policy).

x/kafka/consumer/consumer.go
joshjon: Well the context cancel is usually invoked via a `defer`, so handling SIGINT and SIGTERM would eventually result in the context being cancelled too.

x/kafka/consumer/consumer.go
amatueva: 👍 Maybe worth leaving a comment/adding to a doc that user should implement this behaviour in handler function?

> Publishing to a dead letter queue is supported with the default behaviour of auto committing offsets on read, from which the message can be sent to a dead letter queue if it cannot be handled.

Does Segment have out of the box implementation for this? I'm not familiar with Segment Kafka library, could you point me to where it is implemented?

x/kafka/consumer/consumer.go
amatueva: What I'm trying to say that you have a select statement on ctx.Done() that will fire on context cancellation (which will happen during redeploy, right?) so returning ctx.Error() will result in every deploy polluting error logs?

amatueva: Maybe I should zoom with ya, my brain is confused after all of the onboarding stuff 😄

x/kafka/consumer/consumer.go
joshjon: Added the doc in this commit: https://github.com/cultureamp/ca-go/pull/83/commits/158aa1e1dd5a82cea3c52647786d85847cb7caf7

> Does Segment have out of the box implementation for this?

No it doesn't unfortunately

x/kafka/consumer/consumer.go
joshjon: Resolving this convo based on our zoom chat

x/kafka/consumer/consumer.go
amatueva: Discussed offline that users of the library user should return nil if unretryable error/implement happened so we won't get infinite retries here

Allen-Ning (approved):

amatueva (approved):
LGTM!

PR #89, opened 2022-12-21
chore(decrypter/consumer) fix minor issues

joshjon:
This PR makes the following changes:

- Removes duplicate group name in consumer ID
- Removes decrypt error log from vault decrypter. This was causing a lot of unnecessary noise for "key not found' errors, which are expected when users have been erased.


joshjon:
Does anyone know how to reduce the strictness of coveralls? Its pretty ridiculous that it fails due to reduced test coverage when the changes don't even affect any tests.

hughevans (approved):

PR #94, opened 2023-01-16
[COSE-901] Update LD wrapper to use Contexts

El-Wel:
## Purpose 
LD has updated their SDKs to use contexts rather than users, this has many positive implications for CA including being able to target based on subdomain without needing to use segments. Firstly, wrappers must be updated to use Contexts.

## Context
See [here](https://cultureamp.atlassian.net/wiki/spaces/CoSe/pages/3016884524/Contexts+Update+Strategy) for more context on the context upgrade strategy


x/launchdarkly/flags/config.go
```diff
@@ -154,7 +156,17 @@ func configForLambdaMode(env configurationJSON, cfg *LambdaModeConfig) ld.Config
 
 	// Set the Dynamo base URL if one was provided explicitly.
 	if cfg != nil && cfg.DynamoBaseURL != "" {
-		datastoreBuilder.ClientConfig(aws.NewConfig().WithEndpoint(cfg.DynamoBaseURL))
+		customResolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
```
El-Wel, line 159: test this using `go get github.com/cultureamp/ca-go@<d24199f89ae91da4ff94b5b8bfbeb4e237de89d3>`

x/launchdarkly/flags/client_test.go
```diff
@@ -145,8 +143,7 @@ func assertTestJSONFlags(t *testing.T, c *Client) {
 
 func TestClientLambdaMode(t *testing.T) {
 	t.Run("configures for Lambda (daemon) mode", func(t *testing.T) {
-		os.Setenv(configurationEnvVar, validConfigJSON)
-		defer os.Unsetenv(configurationEnvVar)
```
einal3m, line 149: I assume the `t` version of `Setenv` is helpful and does the unsetting at the end?

x/launchdarkly/flags/client_test.go
```diff
@@ -158,7 +155,7 @@ func TestClientLambdaMode(t *testing.T) {
 	})
 
 	t.Run("configures for Lambda mode with optional overrides", func(t *testing.T) {
-		os.Setenv(configurationEnvVar, validConfigJSON)
+		t.Setenv(configurationEnvVar, validConfigJSON)
 		defer os.Unsetenv(configurationEnvVar)
```
einal3m, line 159: you can probably remove this `Unsetenv` now?

x/launchdarkly/flags/evaluationcontext/account.go
```diff
@@ -0,0 +1,117 @@
...
+	return func(a *Account) {
+		a.subdomain = subdomain
+	}
+}
+
+// NewAnonymousAccount returns a context object suitable for use in unauthenticated
+// requests or requests with no access to user identifiers.
+// Provide a unique session or request identifier as the key if possible. If the
+// key is empty, it will default to an uuid so percentage rollouts will still apply.
+func NewAnonymousAccount(key string) Account {
```
einal3m, line 56: When would you use `NewAnonymousAccount` over `NewAnonymousUser`? I guess the targeting would be slightly different for those two contexts? 🤔

x/launchdarkly/flags/evaluationcontext/account.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/cultureamp/ca-go/x/request"
+	"github.com/google/uuid"
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
+)
+
+const (
+	accountAttributeAccountID = "accountID"
+)
+
+// User is a type of context, representing the identifiers and attributes of
```
einal3m, line 16: You've left a few "user" comments in this file.

einal3m (commented):
This looks great... just a few very minor things... the questions are just for my curiosity 😂

x/launchdarkly/flags/client_test.go
El-Wel: yes! very handy function of the testing library that cleans up after each test

x/launchdarkly/flags/client_test.go
El-Wel: Oop good catch! Missed one

x/launchdarkly/flags/evaluationcontext/account.go
El-Wel: The targeting is different for the two contexts, though only in terms of what `kind` you decide to target on. I am not sure if there really is a need for both, essentially they are the same. 

Using anonymous contexts means they are not stored in LD (won't show up in autofill or in the contexts list) so if dummy accounts were to be used or we don't care on what the user is (more or less if we aren't targeting individually) then you would want them to be anonymous. 

My thoughts with having this option of `NewAnonymousAccount` would be to indicate that the release would be based on account rather than user but in practice they aren't different aside from selecting the kind to target on. What are your thoughts on this? Would it simplify it to only have one anonymous type (user)?

x/launchdarkly/flags/evaluationcontext/account.go
El-Wel: 🤦‍♀️ Will do a more thorough search, dang comments always catch me

x/launchdarkly/flags/evaluationcontext/account.go
einal3m: Nah, I reckon it's useful to have both... 👍

einal3m (approved):

x/launchdarkly/flags/config.go
```diff
@@ -154,7 +156,17 @@ func configForLambdaMode(env configurationJSON, cfg *LambdaModeConfig) ld.Config
...
+			return aws.Endpoint{
+				URL: cfg.DynamoBaseURL,
+			}, nil
+		})
+
+		defaultcfg, err := config.LoadDefaultConfig(context.TODO(), config.WithEndpointResolverWithOptions(customResolver))
```
andrewcameronsims, line 165: I think we may need to replace this `TODO` context with, e.g., the background context, or pass down a context from higher up in the call stack if there is one.

El-Wel:
Tested new wrapper in [feature flag example app](https://github.com/cultureamp/feature-flag-example-app/pull/15) on ibex. Connected fine with Lambda mode without additional config, would not connect with DynamoBaseURL (using aws [endpoints](https://docs.aws.amazon.com/general/latest/gr/ddb.html)) and timed out with the new wrapper but was closed if using the old version of the wrapper. Either there is an issue with the endpoint value that was passed in or this feature never worked 🤔 
Note: when using the old wrapper the lambda did finish execution while the new one did not (timed out)

New wrapper [logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fff-example-app-ibex-add/log-events/2023$252F01$252F27$252F$255B$2524LATEST$255D42c3df58614943a2b0a013576b5f8c58)
Screen Shot 2023-02-01 at 10 29 24 am

Old wrapper [logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fff-example-app-ibex-add/log-events/2023$252F01$252F31$252F$255B$2524LATEST$255D09a949744f2c49c1b00c70e1c6e55192)
Screen Shot 2023-02-01 at 10 30 08 am

PR #95, opened 2023-01-26
fix: handle empty Vault response

andrewcameronsims:
## Background

This is a fix for a panic that occurred in Sentry.

Some investigation revealed that this is probably caused by an edge case we haven't handled where the Vault server responds with a 404 status code and an empty body. In that case a `(nil, nil)` set of return values will be passed up in the library code until it returns in our calling code. Since we're only checking the error, the nil pointer will be dereferenced which causes a panic.

Since this panic happened during a likely leader election in our HA Vault cluster, I think that's _probably_ what happened. But whether or not it is, we'll need to handle this edge case.

## What this PR changes

* Add explicit field names for the test cases in the table-driven tests for `vault_decrypter`. I was getting confused trying to figure out which field was which.
* Add test case to model the scenario where Vault returns an empty body
* Add code to handle this scenario

## Refs

https://github.com/hashicorp/vault/issues/18836

### DIFF
--- a/x/vault/vault_decrypter.go
+++ b/x/vault/vault_decrypter.go
@@ -3,6 +3,7 @@ package vault
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


El-Wel (approved):
Great job, thanks for making the tests clearer too!

PR #96, opened 2023-02-08
Add kafkatest package  and consumer integration tests 

joshjon:
This PR adds package  `kafkatest` which contains a Kafka test client that allows you to easily setup a test topic to interact with in integration tests.

The PR also adds integration tests to the `consumer` package which is used to test consumer integration with a local instance of Kafka.


amatueva (approved):
LGTM!

PR #97, opened 2023-02-08
Add consumer batching with concurrency

joshjon:
## 💬 Description

Add batch processing with concurrency to consumer in order to improve performance, especially on a single partition basis.

The integration tests have been updated to log performance results. Looking at the results it is clear that batching greatly increases performance when consuming messages from a single partition. 

Take for example the results for consumers with a message handler hardcoded to sleep for 50ms:

| **Consumers** | **Partitions** | **Messages Processed** | **Batching** | **Time Taken** | **Messages Per Second** |
|---------------|----------------|------------------------|--------------|----------------|-------------------------|
| 1             | 1              | 50                     | false        | 2.82s          | 17.7                    |
| 10            | 10             | 1000                   | false        | 5.65s          | 177.0                   |
| 1             | 1              | 1000                   | true         | 2.61s          | 382.2                   |
| 10            | 10             | 1000                   | true         | 2.57s          | 389.2                   |

For a single consumer and partition, batching had a 2059.32% increase in performance when compared to a run without batching. 

This improvement aims to help eleviate the pain of our account based partitioning strategy on the Employee Data Topic.  For instance, if a bulk user import occurs for an account, all of the generated events are published to the same partition, which currently results in sequential consumption regardless of the number of consumers in the group. However, with batching, messages on the same partition from a bulk import can be processed concurrently.

## ☘️ Proposed Changes

- Add consumer option `WithMessageBatching` that accepts `batchSize` and `getOrderingKeyFn`.
    - `batchSize` is an `int` which specifies how many messages should be read and processed within a single batch.
    - `getOrderingKeyFn`is a func with the signature`func(message kafka.Message) (string, error)`. It defines how to retrieve the ordering key in which each message within the batch will be stored under.
- Add method `Consumer.processBatch`.
    - Reads in a batch of messages as defined by the batch size.
    - Stores messages in a map of type `map[string][]kafka.Message`, where the key is generated by `getOrderingKeyFn`
    - Spawn a new goroutine for each `key` in the map and handle them in order
- Add unit tests for consumer batching
- Add integration tests for consumer batching


andrewcameronsims (approved):
Awesome PR, A++ would approve again 💯

amatueva (approved):
LGTM! Love the comments and PR description

PR #98, opened 2023-02-10
Fix consumer goroutine data race

joshjon:
## 💬 Description

This PR resolves 2 goroutine data races found in the `consumer` package. 

1. Consumer field `close` is accessed by multiple goroutines when using a `Group`.
2. Unit tests increment a counter from multiple goroutines.

Both of these are of low impact since the the first is only relevant when a Group is being stopped and the second is only only applies to a few unit tests. 

Whats interesting is that an atomic increment is currently being used in the unit tests but this still did not pass the data race check. On the other hand, changing to a mutex based counter did pass the race test.

☘️ Proposed Changes

- Use channels to communicate when a consumer is stopped instead of accessing a shared variable.
- Add type `mutexCounter` to tests which allows a counter to be safely incremented and accessed across multiple goroutines.
- Increase schema registry health check retry count in docker compose.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ test:
 .PHONY: test-ci
 test-ci:
 	mkdir artifacts
-	go test ./... -covermode=atomic -coverprofile=artifacts/count.out
+	go test -race ./... -covermode=atomic -coverprofile=artifacts/count.out
 	go tool cover -func=artifacts/count.out | tee artifacts/coverage.out
 
 # ensures that `go mod tidy` has been run after any dependency changes
--- a/x/kafka/consumer/consumer_test.go
+++ b/x/kafka/consumer/consumer_test.go
@@ -8,7 +8,6 @@ import (
 	"math/rand"
 	"strconv"
 	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ test:
 .PHONY: test-ci
 test-ci:
 	mkdir artifacts
-	go test ./... -covermode=atomic -coverprofile=artifacts/count.out
+	go test -race ./... -covermode=atomic -coverprofile=artifacts/count.out
```
andrewcameronsims, line 12: 💯

x/kafka/consumer/consumer_test.go
```diff
@@ -489,3 +483,20 @@ func (b *testBackoff) NextBackOff() time.Duration {
...
+	m.Lock()
+	defer m.Unlock()
+	m.count++
+}
+
+func (m *mutexCounter) Get() int {
+	m.Lock()
+	defer m.Unlock()
+	return m.count
+}
```
andrewcameronsims, line 502: What happens if we have concurrent reads of the counter, is that similarly unsafe? Not that we really care that much about optimising these tests, but I was just wondering whether a `RWMutex` is suited to this use case.

andrewcameronsims (dismissed):
Awesome work

x/kafka/consumer/consumer_test.go
amatueva: Oh, one question - is pointer to struct WITH mutex is enough? Should it be pointer to mutex itself? 🤔

joshjon: @andrewcameronsims good callout, `RWMutex` is better here since we are doing concurrent reads from the different consumers. I have pushed up a commit to change to `RWMutex`

joshjon: > Should it be pointer to mutex itself

@amatueva All the examples of mutexes I've come across in the past have not used a pointer on the mutex. I don't believe it makes a difference. The data race test also passes with the current non pointer implementation. I think we are all good 👍

amatueva: Ah, pointers on mutexes is a must but since there's a pointer on the struct it should be sufficient!

amatueva (approved):
LGTM! ❤️

andrewcameronsims (approved):

PR #104, opened 2023-02-15
Fix consumer stalling when fetching messages for batch

joshjon:
## 💬 Description

The current implementation of batching improves performance on a single partition, which is particularly useful for unbalanced topic partitions. However, there is a small flaw in how messages are fetched for the batch. The logic requires all messages to first be retrieved for the batch, followed by the concurrent handling of the messages, and finally committing all messages once all handling is complete. But this results in the consumer getting stuck when there isn't enough messages available on the topic for the consumer to fetch for the next batch.

For example, if the batch size is 30 but only 12 messages are available on the topic, then the consumer will fetch those 12 messages and keep waiting to fetch more messages until the batch size of 30 is reached. This is not ideal because it means we will always have a small amount of consumer lag.

To resolve the issue, we should instead stop fetching early if it is clear that there are no new messages to fetch for the batch. With that said, this PR contains such a solution which involves asynchronously fetching messages and concurrently handling them, but will terminate the fetching early if there are no new messages available and the amount of handled messages equals the number of already fetched messages.

## ☘️ Proposed Changes

- Add type `batchProcessor` which is responsible for all batch processing within the consumer. It includes mechanisms to asynchronously fetch and process messages from a Kafka topic within a batch.
- Add type `handlerExecutor` which is responsible for executing the provided handler function on the message. This was required to easily share the handler execution logic between the batch processor and consumer.


hughevans (approved):

x/kafka/consumer/consumer.go
```diff
@@ -418,3 +310,20 @@ func NonStopExponentialBackOff() backoff.BackOff { //nolint:ireturn
...
+	m.Lock()
+	defer m.Unlock()
+	m.v++
+}
+
+func (m *safeCounter) val() int {
+	m.RLock()
+	defer m.RUnlock()
+	return m.v
+}
```
andrewcameronsims, line 329: Not in scope, but wondering if there might be a more appropriate home for inventions like these within the larger `ca-go` library.

andrewcameronsims (approved):
Once again very instructive PR, amazing work 🎉 

Just thinking it over and I know that on the Kafka producer side they have a setting called `linger.ms` which is meant to address this same sort of problem. It's an upper time limit that the producer will wait for a batch to fill up. After that it just sends the batch even if it's undersize, but I think it's supposed to allow you to optimise for your particular expected throughput.

For example I can imagine we get into situations where in our solution we keep processing undersize batches due to the way new arrivals in the queue are distributed. Maybe later on we can consider a similar sort of configurable solution.

joshjon:
@andrewcameronsims I hadn't heard of `linger.ms` before, thanks for sharing! I think it would definitely be great to explore this at some stage. I feel like there is quite a lot of features directly on Kafka that we can make use of which we just aren't potentially aware of, this being one of them. Once our consumers are in a steady state it would be great to look at such features for further optimisations.

joshjon:
@andrewcameronsims Also I just realised something worth mentioning -- `kafka-go` doesn't actually treat 'batching' as we would normally expect. When we we use the Reader to fetch messages, behind the scenes it fetches from Kafka in batches, but it then sends them individually to a channel. This channel is used in the `FetchMessage` method to receive the next message from the channel. The main flaw in this design is that it doesn't actually allow something like `linger.ms` to resolve the particular issue in this PR because there's no way for our consumer to actually know when a batch of messages has been fetched.

So we unfortunately have to work with individual messages received from a channel and group them into 'batches', which is seperate from the batching used by the reader under the hood.

andrewcameronsims:
I was not suggesting that we rely on Kafka to resolve this issue. That is a producer-side config anyway, and I don't think it exists on the consumer side. I was just pointing to the general pattern as something we might think about implementing in our own batching mechanism here at some point in the future.

joshjon:
Ah right sorry there was a bit of confusion on my end about. Makes sense 🙂

x/kafka/consumer/consumer.go
joshjon: Yea agreed. If we need to leverage more concurrency based code then it could be worth including it in ca-go.

PR #122, opened 2023-02-19
Add batchId to consumer debug logs

joshjon:
Add batchId to consumer debug logs to easily track logs for a particular consumer batch.


kings-wang (approved):
LGTM

PR #124, opened 2023-02-20
Fix consumer batching CPU usage

joshjon:
The consumer batching early stop condition in the processing loop is currently very inefficient. The stop condition is a sub loop that loops indefinitely whilst locking and unlocking a sync mutex until the stop condition is reached.

A CPU profile test of the consumer batching showed that the stop condition loop used  ~96% of the CPU.

image

This PR changes the stop condition loop so that it blocks until a new message is available or all current messages have been handled, in which the batch can be stopped early.

This change completely removes the extreme CPU usage.

image

### DIFF
--- a/x/kafka/consumer/consumer_test.go
+++ b/x/kafka/consumer/consumer_test.go
@@ -104,7 +104,7 @@ func TestConsumer_Run(t *testing.T) {
 		require.Equal(t, currMsg, msg.Message)
 		i++
 		if i == wantTimes {
-			consumer.Stop()
+			require.NoError(t, consumer.Stop())
 		}
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


x/kafka/consumer/consumer_test.go
```diff
@@ -477,3 +477,20 @@ func (b *testBackoff) NextBackOff() time.Duration {
 	}
 	return 0
 }
+
+type safeCounter struct {
```
joshjon, line 481: Just moved this inside the test file because its no longer used by the actual consumer

x/kafka/kafkatest/test_client.go
```diff
@@ -162,3 +162,8 @@ func (c *TestClient[EventType]) Writer() *kafka.Writer {
 func (c *TestClient[EventType]) Reader() *kafka.Reader {
 	return c.reader
 }
+
+// Registry returns the internal kafka test registry.
+func (c *TestClient[EventType]) Registry() *TestRegistry[EventType] {
```
joshjon, line 167: Exposes the registry for test purposes

amatueva (approved):

PR #125, opened 2023-02-22
User timer for consumer batching early stop condition

joshjon:
## Description

Currently the consumer batch stop condition is done in a manner that aimed to stop the batch in the shortest time possible. This requires a lot of async coordination between the batch fetcher and processor which has ultimately led to intermittent race conditions causing deadlocks. It is possible to fix this within the current early stop condition, however, it would likely result in more cumbersome code all for minimal gains.

The solution to this problem is to simplify the over-engineered code that currently exists. This is achieved by simply using a timer to stop the batch if the batch is not filled to the specified batch size. This will result in the exact same performance for high volumes and will only slightly impact a scenario were consumed events are received in small increments, which is a non issue in the grand scheme of things.

### DIFF
--- a/x/kafka/consumer/batching.go
+++ b/x/kafka/consumer/batching.go
@@ -4,18 +4,29 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/segmentio/kafka-go"
 	"golang.org/x/sync/errgroup"
 )
 
+type batchProcessorConfig struct {
+	consumerID       string
+	batchSize        int
+	fetchDuration    time.Duration
+	debugLogger      DebugLogger
+	getOrderingKeyFn GetOrderingKey
+	handlerExecutor  *handlerExecutor
+	reader           Reader
+}
+
 type batchProcessor struct {
 	consumerID       string
 	batchSize        int
 	handlerExecutor  *handlerExecutor
 	reader           Reader
+	fetchTimeout     time.Duration
 	fetched          chan kafka.Message
 	processed        chan kafka.Message
 	stop             chan struct{}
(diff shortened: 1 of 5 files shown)


x/kafka/consumer/consumer.go
```diff
@@ -84,6 +90,18 @@ func NewConsumer(dialer *kafka.Dialer, config Config, opts ...Option) *Consumer
...
 		config.ID = uuid.New().String()
 	}
+	if config.MaxBytes == 0 {
+		config.MaxBytes = 1e6 // 1 MB
+	}
+	if config.MaxBytes == 0 {
+		config.MaxBytes = 10e6 // 10 MB
+	}
+	if config.MaxWait == 0 {
+		config.MaxWait = 250 * time.Second
```
amatueva, line 100: Milliseconds?

joshjon: Omg lol. Will fix

x/kafka/consumer/batching.go
```diff
@@ -25,25 +36,29 @@ type batchProcessor struct {
 	debugKeyVals     []any
 }
 
-func newBatchProcessor(consumerID string, debugLogger DebugLogger, reader Reader, handlerExecutor *handlerExecutor, getOrderingKeyFn GetOrderingKey, batchSize int) *batchProcessor {
+func newBatchProcessor(config batchProcessorConfig) *batchProcessor {
+	if config.fetchDuration == time.Duration(0) {
+		config.fetchDuration = time.Second * 10
```
amatueva, line 41: Should we maybe set it to what kafka [default](https://www.oreilly.com/library/view/kafka-the-definitive/9781491936153/ch04.html#:~:text=fetch.max.wait.ms&text=By%20default%2C%20Kafka%20will%20wait,amount%20of%20data%20to%20return.) is?

amatueva (approved):

amatueva (approved):

x/kafka/consumer/batching.go
joshjon: As discussed, `config.fetchDuration` is used in relation to how long we should fetch from the kafka reader's internal message channel. This is unrelated to the actual Kafka client making requests to receive batches.

As such, I think it makes sense to keep this value higher than the one used by the client.

PR #200, opened 2023-09-14
fix: prebuild so that linting is faster

andrewcameronsims:
We're getting tons of timeouts in the linting stage of our CI.

Since building takes the most time in the linter, doing that beforehand seems to prevent linter timeout.

[COSE-520](https://cultureamp.atlassian.net/browse/COSE-520)

[COSE-520]: https://cultureamp.atlassian.net/browse/COSE-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -35,11 +35,14 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Build
+        run: go build ./...
+
       - name: Lint
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.51.2
-          args: "-v --verbose --timeout=5m"
+          args: "-v --timeout=5m"
 
   go-test:
     runs-on: ubuntu-latest


.github/workflows/pipeline.yml
```diff
@@ -35,11 +35,14 @@ jobs:
...
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Build
+        run: go build ./...
+
       - name: Lint
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.51.2
-          args: "-v --verbose --timeout=5m"
```
einal3m, line 42: I think `-v` and `--verbose` are the same thing 🤔

andrewcameronsims: Yep, that's why I removed it, check out line 45.

slacis (approved):

einal3m (approved):

PR #207, opened 2023-09-19
move env from Glamplify to ca-go

Allen-Ning:
This is copy-and-paste work from Glamplify to ca-go for env package

### Context
The original code is [here](https://github.com/cultureamp/glamplify/tree/master/env).

My assumption is that
- [x]  we will dump overlapped package like `log` in Glamplify but not for env package
- [x] there is no clear domain owner for env package so that dropping the code into ca-go instead creating ca-env-go

Also it seems like env package is widely used within CA - check [here](https://github.com/search?q=org%3Acultureamp%20github.com%2Fcultureamp%2Fglamplify%2Fenv&type=code)


andrewcameronsims (dismissed):

AndrewMcCraeCA:
Is it ok if I update this to not be in the `x/` folder and then merge this @Allen-Ning ?

Allen-Ning:
> Is it ok if I update this to not be in the `x/` folder and then merge this @Allen-Ning ?

Sure thing. I'm hesitant to merge this pr into `/x` folder. Feel free to make the change needed.

techmanmike (approved):

PR #280, opened 2024-01-03
playing around with a possible CA logger impl

techmanmike:
Playing around with a simple singleton logger impl


log/logger_example_test.go
```diff
@@ -0,0 +1,38 @@
...
+		Farm:       "test-farm",
+	})
+
+	err := errors.New("exception")
+	logger.Error(ctx, "Event Details", err).
+		Str("test-str", "str").
+		Int("test-number", 1).
+		Msg("initialise handler")
+
+	// Output:
```
AndrewMcCraeCA, line 36: Think somehting is up with the Debug method, we should see this fail as we would expect output to be:
```
{"level":"debug","app":"test-app","app_version":"1.0.0","aws_region":"us-east-1","aws_account_id":"","farm":"test-farm","event":"event_details","test-str":"str","test-number":1,"message":"initialise handler"}
{"level":"error","app":"test-app","app_version":"1.0.0","aws_region":"us-east-1","aws_account_id":"","farm":"test-farm","error":"exception","event":"event_details","test-str":"str","test-number":1,"message":"initialise handler"}
```

AndrewMcCraeCA: This was due to the default logging level being Info, so can be ignored

log/config_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewLoggerConfigWithNoEnv(t *testing.T) {
+	unsetEnvironmentVariables()
+	defer unsetEnvironmentVariables()
```
AndrewMcCraeCA, line 12: do we need to have this deferred call? I don't think these get set any where in this test

log/logger.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// Info starts a new message with info level.
+//
+// You must call Msg or Send on the returned event in order to send the event to the output.
+func Info(ctx context.Context, event string) *zerolog.Event {
+	return defaultLogger.Info(ctx, event)
+}
+
+func (l *Logger) Info(ctx context.Context, event string) *zerolog.Event {
```
AndrewMcCraeCA, line 97: (l *Logger) Info, Warn, Error, Panic, Fatal can all be private now?

log/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	unsetEnvironmentVariables()
+
+	config := newLoggerConfig()
+	assert.NotNil(t, config)
+	assert.Equal(t, "", config.AppName)
+	assert.Equal(t, "1.0.0", config.AppVersion)
+	assert.Equal(t, "", config.AwsRegion)
+	assert.Equal(t, "INFO", config.LogLevel)
+	assert.Equal(t, "local", config.AwsAccountID)
+	assert.Equal(t, "local", config.Farm)
```
AndrewMcCraeCA, line 20: minor thing, but think we're missing checking config.Product here and in the test below

AndrewMcCraeCA (approved):

PR #283, opened 2024-01-05
made all settings methods package scope for easy use/access

techmanmike:
Move Settings to be private and exposed all the env vars a package level scoped functions.

This means clients don't have to pass around a *Settings or manage the creation of calling NewSettings() etc.


env/settings.go
```diff
@@ -1,106 +1,240 @@
...
+// IsProduction returns true if "APP_ENV" == "production".
+func IsProduction() bool {
+	return defaultSettings.isProduction()
+}
+
+func (s *Settings) isProduction() bool {
+	return s.AppEnv == "production"
+}
+
+// IsRunningInAWS returns true if "APP_ENV" == "production".
```
AndrewMcCraeCA, line 224: This is actually more like if "APP_ENV" != "local"?

env/settings.go
```diff
@@ -1,106 +1,240 @@
...
+func IsRunningInAWS() bool {
```
AndrewMcCraeCA, line 225: I'm not sure we will actually need this, as you could just use if !IsRunningLocal() if you ever need to do something different for code in aws but not too fussed about keeping it

AndrewMcCraeCA (dismissed):

env/settings.go
AndrewMcCraeCA: I just updated it

AndrewMcCraeCA (approved):

PR #284, opened 2024-01-05
added in secrets to stable packages

techmanmike:
Added secrets packages that wraps AWS Secret Manager with methods:
- Get

### DIFF
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -1,0 +1,58 @@
+package secrets
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+)
+
+// smClient used for mock testing.
+type smClient interface {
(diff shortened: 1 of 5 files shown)


secrets/secrets.go
```diff
@@ -0,0 +1,59 @@
...
+// AWSSecrets supports the GetSecretValue method.
+type AWSSecrets struct {
+	client smClient
+}
+
+var defaultAWSSecrets = getInstance()
+
+func getInstance() *AWSSecrets {
+	// Should this take dependency on 'env' package and call env.AwsRegion()?
+	region := os.Getenv("AWS_REGION")
```
AndrewMcCraeCA, line 26: Don't think we should use the env package here, so as we can keep the different packages separate.

AndrewMcCraeCA: Curious as to what will happen when AWS_REGION is not set.
It will default to "" but nothing will fail until line 31?
Is there somewhere we can throw an error?
Maybe we should add a default to us-west-2 since odds are this will be the env?

AndrewMcCraeCA (approved):

PR #292, opened 2024-01-11
Secrets: added ability to test package level methods

techmanmike:
Added package level methods + tests.
Added the ability of overwrite the default package level implementation + added automatic testrunner that echoes the secret key as the value (does not hit AWS).

### DIFF
--- a/secrets/clients.go
+++ b/secrets/clients.go
@@ -1,6 +1,8 @@
 package secrets
 
 import (
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
@@ -19,16 +21,22 @@ func newSecretManagerClient(region string) *secretsmanager.SecretsManager {
 	return secretsmanager.New(sess)
 }
 
-type testClient struct{}
+type testRunnerClient struct{}
 
-func newTestClient() *testClient {
-	return &testClient{}
+func newTestRunnerClient() *testRunnerClient {
+	return &testRunnerClient{}
 }
 
-func (c *testClient) GetSecretValue(input *secretsmanager.GetSecretValueInput) (*secretsmanager.GetSecretValueOutput, error) {
-	retVal := &secretsmanager.GetSecretValueOutput{
-		SecretString: input.SecretId,
-	}
+func (c *testRunnerClient) GetSecretValue(input *secretsmanager.GetSecretValueInput) (*secretsmanager.GetSecretValueOutput, error) {
+	arn := "arn:aws:secretmanager:eu-west-2:abc123:secret/id"
+	now := time.Now()
 
+	retVal := &secretsmanager.GetSecretValueOutput{
+		ARN:          &arn,
+		CreatedDate:  &now,
+		Name:         input.SecretId,
+		SecretString: input.SecretId, // just echo back the key as the secret when running in a test
+		VersionId:    input.VersionId,
+	}
 	return retVal, nil
 }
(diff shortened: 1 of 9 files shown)


secrets/clients.go
```diff
@@ -0,0 +1,42 @@
...
+	return secretsmanager.New(sess)
+}
+
+type testRunnerClient struct{}
+
+func newTestRunnerClient() *testRunnerClient {
+	return &testRunnerClient{}
+}
+
+func (c *testRunnerClient) GetSecretValue(input *secretsmanager.GetSecretValueInput) (*secretsmanager.GetSecretValueOutput, error) {
```
AndrewMcCraeCA, line 30: could we make the signature for GetSecretValue( input string)( string,error)?
I think this would make mocking easier in test as your test won't need to import aws packages.

secrets/clients.go
```diff
@@ -0,0 +1,42 @@
+package secrets
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
```
AndrewMcCraeCA, line 6: maybe we should move to github.com/aws/aws-sdk-go-v2 ?
think this changes the secretmanager client construction though.
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/secretsmanager

techmanmike: Done.

secrets/clients.go
techmanmike: Done

secrets/secrets.go
```diff
@@ -1,59 +1,45 @@
...
-
-var defaultAWSSecrets = getInstance()
-
-func getInstance() *AWSSecrets {
-	// Should this take dependency on 'env' package and call env.AwsRegion()?
-	region := os.Getenv("AWS_REGION")
-	return NewAWSSecrets(region)
+// NewAWSSecretsManager creates a new AWS Secret Manager for a given region.
+func NewAWSSecretsManager(region string) (*AWSSecretsManager, error) {
+	// Should this be passed in?
```
AndrewMcCraeCA, line 17: I don't know if this would make sense but also could this context be part of SecretsManager or the client so as to be used when we call GetSecretValue so as to not need to pass in context to Get?

secrets/package.go
```diff
@@ -0,0 +1,50 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/config"
+)
+
+// DefaultAWSSecrets is a public *AWSSecretsManager used for package level methods.
+var DefaultAWSSecrets = getInstance()
```
AndrewMcCraeCA, line 14: maybe this should be DefaultAWSSecretsManager?

secrets/secrets.go
techmanmike: I've added it to parameters - as it might be useful for Deadline or other things...

secrets/package.go
techmanmike: fixed

secrets/secrets.go
```diff
@@ -1,59 +1,43 @@
...
-		client: client,
+// NewAWSSecretsManagerWithClient creates a new AWS Secret Manager with a custom client
+// that supports the AWSSecretsManagerClient interface.
+func NewAWSSecretsManagerWithClient(client AWSSecretsManagerClient) *AWSSecretsManager {
+	return &AWSSecretsManager{
+		Client: client,
 	}
 }
 
 // Get retrives the secret from AWS SecretsManager.
```
AndrewMcCraeCA, line 33: typo: retrieves

AndrewMcCraeCA (approved):

PR #294, opened 2024-01-17
Significant updates to logging

techmanmike:
Added:
- a test impl that does a noop if running in a test
- expose the NewX methods implementations so that people can manage their own
- a conversion method for legacy glamplify logging from log.fields


log/constants.go
```diff
@@ -0,0 +1,22 @@
+package log
+
+const (
```
AndrewMcCraeCA, line 3: I'm not totally against having these in a separate file but since these are any of these ever used outside of config? I feel like they maybe should just live theyr'e unless used elsewhere?

log/constants.go
```diff
@@ -0,0 +1,22 @@
...
+
+	// *** Global Environment Variables ***.
+	AppNameEnv       = "APP"
```
AndrewMcCraeCA, line 6: also since these are in the package, and the tests are also in the package, do these need to be Public?
Unless we intend to people to consumers of log to use these?

log/LOGGER.md
```diff
@@ -15,97 +15,186 @@ You MUST set these:
...
+- AWS_ACCOUNT_ID = The AWS account Id this code is running in, defaults to  "development"
 - FARM = The name of the farm or where the code is running, defaults to "local" (eg. "production", "dolly") 
 - APP_VERSION = The version of the application, defaults to "1.0.0"
 
+## Use in Unit Tests
 
-## Examples
+By default the logger will not emit any output when running inside a test. You can override this behaviour by setting the `QUIET_MODE` environment variable to "false".
+
+When running localling you can also set the `CONSOLE_WRITER` to "true" to change from json to key-value colour coded output. Note: Never run with the `CONSOLE_WRITER` set to "true" in production.
```
AndrewMcCraeCA, line 26: typo ?

AndrewMcCraeCA: localling

log/LOGGER.md
```diff
@@ -15,97 +15,186 @@ You MUST set these:
...
+
+	props := SubDoc().
+		Str("str", "value").
+		Int("int", 1).
+		Bool("bool", true).
+		Duration("dur", duration).
+		IPAddr("ipaddr", ipv4).
+		UUID("uuid", u)
+
+	Debug("debug_with_all_field_types").
```
AndrewMcCraeCA, line 61: these might need `log.` in front of them

log/logger.go
```diff
@@ -15,15 +14,7 @@ type Logger struct {
...
-var defaultLogger *Logger = getInstance()
-
-func getInstance() *Logger {
-	setGlobalLogger()
-	config := newLoggerConfig()
-	return newDefaultLogger(config)
-}
-
-func newDefaultLogger(config *LoggerConfig) *Logger {
+func NewLogger(config *Config) *Logger {
```
AndrewMcCraeCA, line 17: Maybe we can add a comment to this to explain the recommended/default approach is not to maintain your own logger.

AndrewMcCraeCA (approved):
This looks good, going to merge as we're migrating some old glamplify packages and want to use this new interface.
we can review changes to this package in future

PR #302, opened 2024-01-25
2671-migrate-aws-package-from-glamplify

morgan121:
## Purpose
We are trying to swith the glamplify aws package to ca-go


techmanmike:
We are not going to copy glamplify code over - and instead ask that Go projects move to the new ca-go interfaces.

PR #306, opened 2024-02-02
Support for Encode & Decode JWTs using JWKs

techmanmike:
Added JWT module to Encode and Decode jwt tokens using RSA keys (only).

The Decoder only support JWKs passed in by string - the web-gateway, perform-core, etc. public keys MUST be in the JWKs JSON file for the Decoder to successfully decode those tokens.

The Encoder only support a single RSA PEM key.


jwt/decoder.go
```diff
@@ -0,0 +1,194 @@
...
+	"encoding/pem"
+	"errors"
+	"fmt"
+
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/lestrrat-go/jwx/jwk"
+)
+
+const (
+	AuthBearerPrefix   = "Bearer "
```
AndrewMcCraeCA, line 16: I can't see this used anywhere, am I missing the use of this?

jwt/jwt_example_test.go
```diff
@@ -0,0 +1,23 @@
+package jwt_test
+
+import (
+	"fmt"
+
+	"github.com/cultureamp/ca-go/jwt"
+)
+
+func BasicExamples() {
```
AndrewMcCraeCA, line 9: I'm not sure if it matters, but think the testable examples uses Example as the prefix so this would be ExampleBasic.
https://go.dev/blog/examples

jwt/decoder.go
```diff
@@ -0,0 +1,194 @@
...
+	KidHeaderKey       = "kid"
```
AndrewMcCraeCA, line 17: I see almost all of these are only used inside the `jwt` package are we making them public for other reasons? or could we make these private? => `kidHeaderKey`?

jwt/encoder_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	encoderAuthKey string = "./testKeys/jwt-rsa256-test-webgateway.key"
```
AndrewMcCraeCA, line 15: i can see this isn't used, but it feels like this could be used in place of `testDefaultAuthPrivateKey` and then make this test file outside the package? `jwt_test` is/was that the intent or should we remove this?

jwt/jwt_example_test.go
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	// Encode this claim with the default "web-gateway" key and add the kid to the token header
+	token, err := jwt.Encode(claims)
+	fmt.Printf("The encoded token is '%s' (err='%v')\n", token, err)
+
+	// Decode it back again using the key that matches the kid header using the default JWKS JSON keys
+	sc, err := jwt.Decode(token)
+	fmt.Printf("The decode token is '%v' (err='%+v')\n", sc, err)
+}
```
AndrewMcCraeCA, line 23: we could also add the example for new encoder/decoder in here?

AndrewMcCraeCA: // To create a specific instance of the encoder and decoder you can use the following
	privateKeyBytes, err := os.ReadFile(filepath.Clean("./testKeys/jwt-rsa256-test-webgateway.key"))
	encoder, err := jwt.NewJwtEncoder(string(privateKeyBytes), jwt.WebGatewayKid)

	claims = &jwt.StandardClaims{
		AccountId:       "abc123",
		RealUserId:      "xyz234",
		EffectiveUserId: "xyz345",
		ExpiresAt:       time.Unix(2211797532, 0), //  2/2/2040
		IssuedAt:        time.Unix(1580608922, 0), // 1/1/2020
		NotBefore:       time.Unix(1580608922, 0), // 1/1/2020
	}

	token, err = encoder.Encode(claims)
	b, err := os.ReadFile(filepath.Clean("./testKeys/development.jwks"))
	decoder, err := jwt.NewJwtDecoderWithDefaultKid(string(b), "web-gateway")

	claim, err := decoder.Decode(token)
	fmt.Printf("The decoded token claim are is '%v' (err='%+v')\n", claim, err)

AndrewMcCraeCA (approved):

AndrewMcCraeCA (approved):

PR #308, opened 2024-02-12
Add Kafka to stable packages

techmanmike:
Added Consume() and Stop() at package level of kafka/consumer/


kafka/consumer/consumer.go
```diff
@@ -0,0 +1,213 @@
...
+	conf               kafka.ReaderConfig
+	reader             Reader
+	withExplicitCommit bool
+	stopCh             chan struct{}
+	clientHandler      *messageHandler
+}
+
+// NewConsumer returns a new Consumer configured with the provided dialer and config.
+func NewConsumer(config Config, opts ...Option) *Consumer {
+	if config.ID == "" {
```
AndrewMcCraeCA, line 68: I wonder if it would be worth creating a NewConfig method to provide these defaults like NewLoggerConfig?
I feel like having allowing these to be passed in empty and creating defaults here is kinda not providing defaults for Config but for Consumers

kafka/consumer/consumer.go
```diff
@@ -0,0 +1,213 @@
...
+type Reader interface {
+	ReadMessage(ctx context.Context) (kafka.Message, error)
+	FetchMessage(ctx context.Context) (kafka.Message, error)
+	CommitMessages(ctx context.Context, msgs ...kafka.Message) error
+	Close() error
+}
+
+// Config is a configuration object used to create a new Consumer.
+type Config struct {
+	ID      string // Default: UUID
```
AndrewMcCraeCA, line 41: I'm not super across the Kafka consumer stuff.
But I don't really see the scenario where you configure the ID yourself?
Could the Consumer.id just be a new UUID every time we create a Consumer
I can see using the Consumer.ID so maybe we could just make that public?

kafka/consumer/consumer.go
AndrewMcCraeCA: We will move this out of x folder and iterate on this.
The config logic can be rewritten

AndrewMcCraeCA (approved):

PR #310, opened 2024-02-16
updated linters

techmanmike:
Updated linter config to include more linters as well as adding a depguard allow/deny list.

Purpose: 

- To guard against adding new dependencies by accident - they now have to be intentional. 
- To help standardize our coding style so that it is more consistent across packages (and later other Go repos)


.golangci.yml
```diff
@@ -109,7 +109,77 @@ linters:
...
+        # Used to determine the package matching priority.
+        # There are three different modes: `original`, `strict`, and `lax`.
+        # Default: "original"
+        list-mode: strict
+        # List of file globs that will match this list of settings to compare against.
+        # Default: $all
+        files:
+          - $all
+        # List of allowed packages.
+        allow:
```
AndrewMcCraeCA, line 142: I know we need to overhaul the readme but maybe we could outline the process/suggested criteria for these allowed or disallowed dependencies somewhere to give later contributors some indication of how they can add new entries or swap existing ones down the track?

techmanmike: Yeah - that is a great idea. Should I put something in the README now? Or wait until we overhaul it?

.golangci.yml
```diff
@@ -109,7 +109,77 @@ linters:
...
+    #  `$gostd` - matches all of go's standard library (Pulled from `GOROOT`)
+    #
+    # Default: Only allow $gostd in all files.
+    rules:
+      # Name of a rule.
+      main:
+        # Used to determine the package matching priority.
+        # There are three different modes: `original`, `strict`, and `lax`.
+        # Default: "original"
+        list-mode: strict
```
andrewcameronsims, line 136: I'm really unclear on what a lot of this config means. Could we link to any documentation in the comments? e.g. how does original differ from strict or lax?

.golangci.yml
```diff
@@ -109,7 +109,77 @@ linters:
...
+          - github.com/rs/zerolog
+          - github.com/segmentio
+          - github.com/lestrrat-go
+          - github.com/golang-jwt
+          # not sure about these below
+          - github.com/IBM/sarama 
+          - github.com/xdg-go/scram
+          - github.com/kelseyhightower/envconfig
+          - github.com/cenkalti/backoff
+        # Packages that are not allowed where the value is a suggestion.
```
andrewcameronsims, line 160: Does a package need to be in the allowlist to be allowed? Or does it just not be in the denylist?

.golangci.yml
```diff
@@ -29,7 +29,7 @@ linters:
     #    - exhaustive
     #    - exhaustruct
     - exportloopref
-    #    - forbidigo
+    # - forbidigo
```
andrewcameronsims, line 32: If we don't want to use this one, should we just remove it from the configuration file?

andrewcameronsims (approved):

.golangci.yml
techmanmike: I've added this to the comment:
# Original is the original way that the package was written to use.  It is not recommended to stay with this and is only here for backwards compatibility.
# Strict, at its roots, is everything is denied unless in allowed.
# Lax, at its roots, is everything is allowed unless it is denied.

.golangci.yml
techmanmike: Answered above

.golangci.yml
techmanmike: To be honest, I'd like to enable more linters but there were so many errors that I couldn't reasonably fix. So maybe we can revisit trying to get ca-go to pass this linter? Thoughts?

andrewcameronsims: What does the linter do exactly? I saw that it forbids certain kinds of identifiers? Couldn't really get a clear idea from the docs.

andrewcameronsims (dismissed):

PR #311, opened 2024-02-20
Pre commit config hooks

techmanmike:
Purpose:
- There have been a number of recent security incidents of keys and other secrets being committed to public repos. This PR adds OPTIONAL support for engineers working with sensitive data or secret keys to be able to detail these before committing to git.
- To be able to turn pre-commit checks ON and OFF depending on the engineer's preferences
- This is NOT intended to be the only gate keeper of detecting secrets, but for those engineers that wish to use this, to make it easy to do so by adding instructions to the README.md and a pre-commit config file with a default set of checkers.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+-   repo: https://github.com/golangci/golangci-lint
+    rev: v1.56.2
+    hooks:
+    -   id: golangci-lint
+-   repo: local
+    hooks:
+      - id: trufflehog
+        name: TruffleHog
+        description: Detect secrets in your data.
+        entry: bash -c 'trufflehog git file://. --since-commit HEAD --only-verified --fail --no-update'
```
benstringer-cultureamp, line 31: Should have a corresponding section in the README saying trufflehog is a pre-req and how to install it.

techmanmike: Sure - will definitely do that if others think this is a good idea to pursue...

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+    -   id: mixed-line-ending
+    -   id: trailing-whitespace
+-   repo: https://github.com/golangci/golangci-lint
+    rev: v1.56.2
+    hooks:
+    -   id: golangci-lint
+-   repo: local
+    hooks:
+      - id: trufflehog
+        name: TruffleHog
```
benstringer-cultureamp, line 29: A nicer alternative may be to use synk here, and configure it to use trufflehog. This means we have one scanning tool (synk) but as it can use trufflehog in the background we get the same result.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+# %> pre-commit install
+# %> pre-commit run --all-files
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0  # Use the ref you want to point at
+    hooks:
+    -   id: check-json
+    -   id: check-yaml
+    -   id: check-added-large-files
+    -   id: detect-aws-credentials
```
dstrates, line 15: I think this and `detect-private-key` will double up on trufflehog's checks

.pre-commit-config.yaml
techmanmike: I've added snyk alongside trufflehog here to show how to make it work.

.pre-commit-config.yaml
techmanmike: Is that a bad idea? As I wonder if they check in different ways etc.

dstrates: Yeah it should be fine - the tradeoff is potentially slower and noisier checks

README.md
```diff
@@ -78,11 +71,38 @@ directory. This works locally and via [GitHub Codespaces](https://github.com/fea
...
 
+### Pre-Commit
+
+This is optional but is recommended for engineers working with highly sensitive secrets or data.
+
+Config for what is checked is stored in `.pre-commit-config.yaml`
+
+Download:
+1. brew install pre-commit
+2. brew install trufflehug
```
dstrates, line 82:
```suggestion
2. brew install trufflehog
```

andrewcameronsims (approved):

PR #312, opened 2024-02-21
Kms client rework

techmanmike:
Removed the automatic magic that detects if running inside a test and swaps out the implementation for a testRunnerClient.

Updated the CIPHER.md and examples.

### DIFF
--- a/cipher/package_test.go
+++ b/cipher/package_test.go
@@ -14,7 +14,7 @@ func TestPackageEncrypt(t *testing.T) {
 
 	// replace the package level client with our mock
 	stdClient := cipher.DefaultKMSCipher.Client
-	cipher.DefaultKMSCipher.Client = newTestRunnerClient()
+	cipher.DefaultKMSCipher.Client = newMockedCipherClient()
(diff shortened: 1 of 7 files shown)


AndrewMcCraeCA (approved):
nice one

cipher/package_test.go
```diff
@@ -12,10 +12,33 @@ func TestPackageEncrypt(t *testing.T) {
 	ctx := context.Background()
 	keyId := "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
 
+	// replace the package level client with our mock
+	stdClient := cipher.DefaultKMSCipher.Client
+	cipher.DefaultKMSCipher.Client = newMockedCipherClient()
+	defer func() {
```
AndrewMcCraeCA, line 18: I'm not sure if we need to do this storing and restoring of the default client in their own tests but think this is fine to leave in

AndrewMcCraeCA (approved):

PR #323, opened 2024-02-27
Support the new JWT Engineering Standard

techmanmike:
Purpose:
- To fully support the recent update to the [JWT engineering standard](https://cultureamp.atlassian.net/wiki/spaces/TV/pages/3253240053/JWT+Authentication)
- Added support for ECDSA keys (preferred) alongside RSA.
- All JWTs MUST have a kid now.
- Enforce 10m Expiry and added clock skew of 10 seconds.


jwt/JWT.md
```diff
@@ -1,16 +1,17 @@
...
+
 - AUTH_PUBLIC_JWK_KEYS = A JSON string containing the well known public keys for Decoding a token.
 - AUTH_PRIVATE_KEY = The private RSA PEM key used for Encoding a token.
 
-You can OPTIONALLY set these:
-- AUTH_PUBLIC_DEFAULT_KEY_ID = The default "kid" to use when no kid in present in the token (default to 'web-gateway`).
-- AUTH_PRIVATE_KEY_ID = The default "kid" to add to the token heading when Encoding.
+You can SHOULD also set:
+
+- AUTH_PRIVATE_KEY_ID = The default "kid" to add to the token heading when Encoding (defaults to "web-gateway")
```
andrewcameronsims, line 14: From my understanding of the standard (which may be wrong), isn't the `kid` a `MUST`?

techmanmike: Yes - you are correct! Let me fix that...

jwt/claims.go
```diff
@@ -83,17 +96,13 @@ func newEncoderClaims(sc *StandardClaims) *encoderStandardClaims {
 		RealUserID:      sc.RealUserId,
 	}
 
+	claims.Issuer = sc.Issuer
+	claims.Subject = sc.Subject
+	claims.Audience = sc.Audience
 	now := time.Now()
-	claims.IssuedAt = claims.correctTime(sc.IssuedAt, now)
+	claims.ExpiresAt = claims.correctTime(sc.ExpiresAt, now.Add(10*time.Minute))
```
AndrewMcCraeCA, line 103: do we need to call out the 10 minute expiry in docs/comments?
should this be configurable at a later point maybe? (more curious if that's approporiate, rather than a suggestion)

AndrewMcCraeCA (approved):

jwt/claims.go
techmanmike: The standard says 10 minutes, but we can make this configurable later if we really need to.

PR #338, opened 2024-03-04
Migrate cultureamp go test coverage

techmanmike:
Purpose:
- Prepare for making ca-go INTERNAL
- Remove coveralls for test coverage reporting
- Replaced with gocover-cobertura and gotestfmt.

To Do:
- Use the https://github.com/cultureamp/go-test-coverage action once we have a Go Proxy and ca-go is made INTERNAL


.coverage-exclusions
```diff
@@ -0,0 +1,2 @@
+# which packages to exclude from coverage calculations
+github.com/cultureamp/ca-go/x/*
```
AndrewMcCraeCA, line 2:
```suggestion
github.com/cultureamp/ca-go/x/*

```

jwt/decoder.go
AndrewMcCraeCA, line 1: Are these changes in the `jwt` package intended to be in this PR? seems like this should be separate from this test coverage PR

techmanmike: Crap - you are right - not sure how that happened! Let me fix...

PR #341, opened 2024-03-06
Logging updates from Golang meetup feedback

techmanmike:
Purpose: Make it easier to migrate from Glamplify logging to CA-GO logging.

Changes:
- Now have new WithGlamplifyRequestFieldsFromCtx(ctx context.Context) method to read values from the context (like glamplify logging does)
- Removed the glamplify logger to encourage migration to the new logging methods
- Changed tests that were NOT assert'ing to be Example() with validated // Output: 
- Can now dynamically set the time func() so that example output is static and UTC


log/config.go
```diff
@@ -77,13 +88,75 @@ func NewLoggerConfig() *Config {
...
+			NoColor:         !c.ConsoleColour,
+			FormatMessage:   c.formatMessage,
+			FormatTimestamp: c.formatTimestamp,
+		}
+	}
+
+	return writer
+}
+
+func (c *Config) formatMessage(i interface{}) string {
```
AndrewMcCraeCA, line 142: do we need to provide our own versions of these?
is there some issue with just using the defaults?
these look to be the defaults: 
https://github.com/rs/zerolog/blob/v1.32.0/console.go#L435C6-L435C33
https://github.com/rs/zerolog/blob/v1.32.0/console.go#L355

techmanmike: Yes - they covert local time which means all the testable Examples fail in CI as my machine in +11:00 vs Github/AWS is UTC.

techmanmike: Basically this is bad in the zerolog code (assuming local):
case string:
			ts, err := time.ParseInLocation(TimeFieldFormat, tt, time.Local)
			if err != nil {
				t = tt
			} else {
				t = ts.Local().Format(timeFormat)
			}

log/extensions_test.go
```diff
@@ -110,97 +108,54 @@ func TestExtensionWithAuthorizationTracing(t *testing.T) {
...
-			str:      "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJhY2NvdW50SWQiOiJkNDc1ZGQ1Yi1mMTZjLTRiZmItODk4Yy1kMzQzNWEyMTUyMzkiLCJlZmZlY3RpdmVVc2VySWQiOiI1YjMxNjY0YS03NjEwLTRmYjAtYmM4OS1mOWY4ZTIwYmY4Y2UiLCJyZWFsVXNlcklkIjoiNWIzMTY2NGEtNzYxMC00ZmIwLWJjODktZjlmOGUrUC021FhB_zuETHmhQUXOfIyTkpvhcJfrrqwdcc-KmJGznckACLj65VmnayoltCce_3JGJ361GuutgrDaqp1aW4D05mvO8CCIRwGq8hTcRoi7IdXYSnA6UlXtLNYvttz92jaAAoNDCZmbbP-umHac4x5AT1xY-kVyh7VAadZG_Qe7dZWU9WCHtCV3mqTMwX9B9zrqY2NrpblevbbYpoiJiXOU7kex4BEivF1K6VWI-mpcmKtEOZLx2E",
-			redacted: "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJhY2NvdW50SWQiOiJkNDc1ZGQ1Yi1mMTZjLTRiZmItODk4Yy1kMzQzNWEyMTUyMzkiLCJlZmZlY3R********************vttz92jaAAoNDCZmbbP-umHac4x5AT1xY-kVyh7VAadZG_Qe7dZWU9WCHtCV3mqTMwX9B9zrqY2NrpblevbbYpoiJiXOU7kex4BEivF1K6VWI-mpcmKtEOZLx2E",
-		},
-	}
-	for _, tC := range testCases {
-		t.Run(tC.desc, func(t *testing.T) {
-			r := redactString(tC.str)
-			assert.Equal(t, tC.redacted, r, tC.desc)
-		})
+func getExampleLoggerConfig(sev string) *log.Config {
```
AndrewMcCraeCA, line 151: we don't ever use any other severity other than "INFO", is it worth just moving this into here and removing the parameter? or were you planning on adding other examples with different severities?

log/package_test.go
```diff
@@ -8,8 +8,29 @@ import (
...
+
+	mockLogger.On("Debug", "should_call_mock").Return(log.SubDoc())
+
+	log.Debug("should_call_mock").
+		Properties(log.SubDoc().
+			Str("resource", "resource_id").
+			Int("test-number", 1),
+		).Details("detailed information explain")
+
+	// Output:
```
AndrewMcCraeCA, line 30:
```suggestion
```

log/package_test.go
```diff
@@ -8,8 +8,29 @@ import (
...
+	//
```
AndrewMcCraeCA, line 31:
```suggestion
```

log/properties.go
```diff
@@ -47,6 +47,13 @@ func (lf *Property) Duration(key string, d time.Duration) *Property {
 	return lf
 }
 
+// Time adds the property key with val as an uuid.UUID to the log.
```
AndrewMcCraeCA, line 50: Had a use for this Time field earlier today, this is a useful addition 👍

log/redact_string_test.go
```diff
@@ -0,0 +1,64 @@
+package log
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+
+func TestStringRedaction(t *testing.T) {
```
AndrewMcCraeCA, line 10: I think it's good to add tests for this but I have some reservations around this redaction logic here and in the JWT PR.

I can't quite describe it but think maybe it's a bit strange that it will behave differently with strings of different lengths?
I'm not sure if I called this redactString method I would be thinking about how long my string was?
Maybe let's not address this in this PR though I'll move this discussion to the JWT PR

log/config.go
AndrewMcCraeCA: ah ok that's ok then, just seeing if we could save some code. but yeh that assumption of always using local by zerolog is annoying, surely they would encounter that on their own local dev vs pipelines for testing too...

log/package_test.go
AndrewMcCraeCA: Do we want an output comment here? or should we remove these?

AndrewMcCraeCA (approved):
some minore suggestions but nothing urgent

log/extensions_test.go
techmanmike: I was leaving that door open to future test writers - yes

log/package_test.go
techmanmike: I was confirming that the mock worked because there was no output.

Or to say it another way: I am asserting there is no output for this test.

So I think that is ok?

log/redact_string_test.go
techmanmike: Agree - I think what we should do is what JT said yesterday "add a interface for redaction to this class so that something outside of the logger does redacting for you".

I can do that in another PR. This one is already pretty big.

PR #342, opened 2024-03-07
refactor ENV package based on Golang meetup feedback

techmanmike:
Purpose: Be able to mock settings and use 3rd party to populate structs from env_vars.

- Add dependency on "github.com/caarlos0/env/v10"
- split Settings struct into Common, AWS, Datadog, etc.
- Dropped CacheTimeout setting (how is the a global setting?)
- Dropped AuthZ settings (given we are still deciding on which tech to use for AuthZ)
- Updated package tests to show how to mock a Setting (if needed)
- Now panic if we can't parse the XYZsettings structs
- APP and AWS_REGION are now required,notEmpty env vars


env/common_settings.go
```diff
@@ -0,0 +1,67 @@
+package env
+
+import (
+	senv "github.com/caarlos0/env/v10"
```
AndrewMcCraeCA, line 4: I guess this doesn't change the public interface, of this env package, but I feel like since kelseyhightower/envconfig is so widely used here at CA I would suggest we go with that rather than a new package?
like there are 8+ repos using it: https://github.com/search?q=org%3Acultureamp+kelseyhightower%2Fenvconfig+language%3AGo&type=code

I think conversations for adopting the ca-go package version would then be easier cause we can just say "it's using envconfig"

env/package.go
```diff
@@ -0,0 +1,249 @@
...
+	return newCommonSettings()
+}
+
+// GetAppName returns the application name from the "APP" environment variable.
+func AppName() string {
+	return DefaultCommonSettings.GetAppName()
+}
+
+// GetAppVersion returns the application version from the "APP_VER" environment variable.
+// Default: "1.0.0".
```
AndrewMcCraeCA, line 40: I like having the default in here in case people need to work out if they are getting the value they expect or the default/falllback.
Is it too much effort to add these just for the package methods?

AndrewMcCraeCA: ah I see plenty more do. Do the ones without this mean there is no one definitive default? like IsProduction?

env/common_settings.go
```diff
@@ -0,0 +1,81 @@
...
+	IsRunningLocal() bool
+}
+
+// commonSettings that drive behavior used by at least 90% of apps.
+type commonSettings struct {
+	// These have to be public so that "github.com/caarlos0/env/v10" can populate them
+	App_Name      string `env:"APP,required,notEmpty"`
+	App_Version   string `env:"APP_VERSION"           envDefault:"1.0.0"`
+	App_Env       string `env:"APP_ENV"               envDefault:"development"`
+	Farm_Location string `env:"FARM"                  envDefault:"local"`
```
AndrewMcCraeCA, line 26: I would maybe go with Farm_Name

AndrewMcCraeCA (approved):
one small suggestion but not a massive blocker

PR #351, opened 2024-03-13
Enable JWT key rotation

techmanmike:
Purpose: Allow for JWT key rotation without having to start/stop services.

Changes:
- NewJwtEncoder/NewJwtDecoder now takes a func() that clients MUST provide that provides a refreshed key/jwks
- Added cache with default 60 minute expiry for the key/jwks
- Added WithExpiryOptions to constructors so that clients can change the 60 minute default
- Package level decoder/encoder provide a func that reads ENV_VARS, so clients using the package level methods just need to update the ENV_VAR to refresh the keys/jwks
- Updated JWT.md


jwt/decoder.go
```diff
@@ -142,6 +167,52 @@ func (d *JwtDecoder) useCorrectPublicKey(token *jwt.Token) (publicKey, error) {
 	return nil, fmt.Errorf("failed to decode: no matching key_id (kid) header for: %s", kid)
 }
 
+func (d *JwtDecoder) getJWKSet() (jwk.Set, error) {
+	// First chech cache, if its there then great, use it!
```
AndrewMcCraeCA, line 171:
```suggestion
	// First check cache, if its there then great, use it!
```

jwt/decoder.go
```diff
@@ -142,6 +167,52 @@ func (d *JwtDecoder) useCorrectPublicKey(token *jwt.Token) (publicKey, error) {
 	return nil, fmt.Errorf("failed to decode: no matching key_id (kid) header for: %s", kid)
 }
 
+func (d *JwtDecoder) getJWKSet() (jwk.Set, error) {
+	// First check cache, if its there then great, use it!
+	obj, found := d.cache.Get(jwksCacheKey)
```
AndrewMcCraeCA, line 172: could this be refactored into a method so it can be reused in refetchJWKSet?
it could just return the jwks.Set, bool
Since I'm not really sure how the cache key can ever point to something that isn't a jwk.set I think we can just return a boolean and not an error?
Isn't the the only place it can be set is from line 220

jwt/encoder.go
```diff
@@ -74,9 +91,89 @@ func (e *JwtEncoder) Encode(claims *StandardClaims) (string, error) {
...
+	if encodingKey.kid != "" {
+		token.Header[kidHeaderKey] = encodingKey.kid
+	}
+
+	return token.SignedString(encodingKey.privateSigningKey)
+}
+
+func (e *JwtEncoder) getPrivateKey() (*encoderPrivateKey, error) {
+	// First chech cache, if its there then great, use it!
+	obj, found := e.cache.Get(privCacheKey)
```
AndrewMcCraeCA, line 103: same as my comment in the decoder I think L103 to L110 can be extracted to it's own method?

jwt/decoder.go
techmanmike: I've done what you suggested @AndrewMcCraeCA - let me know if you like this better?

jwt/encoder.go
techmanmike: I've refactored as you suggested. Thoughts?

jwt/decoder.go
```diff
@@ -5,42 +5,61 @@ import (
...
-	kidHeaderKey         = "kid"
-	algorithmHeaderKey   = "alg"
-	signatureHeaderKey   = "sig"
-	webGatewayKid        = "web-gateway"
-	accountIDClaim       = "accountId"
-	realUserIDClaim      = "realUserId"
-	effectiveUserIDClaim = "effectiveUserId"
+	kidHeaderKey                  = "kid"
+	algorithmHeaderKey            = "alg"
+	signatureHeaderKey            = "sig"
```
AndrewMcCraeCA, line 19:
```suggestion
```
This const is never used anywhere do we need it?

jwt/encoder.go
```diff
@@ -3,56 +3,67 @@ package jwt
...
+	"github.com/patrickmn/go-cache"
 )
 
 const (
-	ecdsaKey512 = iota
+	privCacheKey                  = "encoder_priv_key"
+	defaultEncoderExpiration      = 60 * time.Minute
+	defaultEncoderCleanupInterval = 1 * time.Minute
+
+	invalidKey = iota
```
AndrewMcCraeCA, line 18: maybe I'm not quite across iota and we need this enum for something but invalidKey is not used

jwt/decoder_options_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDecoderOptions(t *testing.T) {
+	b, err := os.ReadFile(filepath.Clean(testAuthJwks))
+	assert.Nil(t, err)
+	validJwks := string(b)
+
+	i := 0
+	jwks := func() string {
```
AndrewMcCraeCA, line 18:
```suggestion
// This will be called each time the cache is refreshed and they we can assert i has been incremented the correct number of times below.
	jwks := func() string {
```
I struggled at first to understand what this test and encoder were doing so thought I'd add a comment?

jwt/decoder_options_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	token := "eyJhbGciOiJSUzUxMiIsImtpZCI6InJzYS01MTIiLCJ0eXAiOiJKV1QifQ.eyJhY2NvdW50SWQiOiJhYmMxMjMiLCJlZmZlY3RpdmVVc2VySWQiOiJ4eXozNDUiLCJyZWFsVXNlcklkIjoieHl6MjM0IiwiaXNzIjoiZW5jb2Rlci1uYW1lIiwic3ViIjoidGVzdCIsImF1ZCI6WyJkZWNvZGVyLW5hbWUiXSwiZXhwIjoyMjExNzk3NTMyLCJuYmYiOjE1ODA2MDg5MjIsImlhdCI6MTU4MDYwODkyMn0.hQLuOe8qZHUgstYe0A0n4-Pww7ovlReyKiDR1Y02ltUnUlgbm9qpp-Ef6YNFuIKdHmS-ynQbDx5pbI36szsggzi80apNpI48cwSXshx82TwuU-_Z4wNBXu7MdPvbA5FdjhxCvRqaqhglsGJ6NofC1bP9awVyyy4j9LGfkVuVEXJQrVpdvEs8Ks-LxlWz7_9Cr7BrZcLuBJnujhe4CbdSudkrfeFl19EY3i1wH9OatGjfjwOSJVqv-ZLnn3QkaZmrQ1xwXTm3MlMUH3KSQjBn8h6vbqosIB5iHDFtqR11mLCgYExGHBpzFjM1d5NEmcTNLV9MtZ_qDZwG0wkgv9O4rXVQ0JfdXypMwhchED2Z45_mc2OiLidtKtDmeoE5g0Daq8YpM0ZpVRbXUFeYIZ1doQKUNsbWNdITmrjVOC3Zn8BecYPu1pC4Hk1y-ViArDzxlCMHA7Bua64BfzVuaJ8pBTEmbqMiZ9VujWcimCOtJ5yfCks_RPAhFYOErcqy3B56fmyYdIN__mKl7VvRDtBSiiPGCq07BUjGywaMoZIULbyXYSV4zs3hX_R4_o4asGiVWCZgn7k4pZzCJo_y2e-Mf85nYoRlyr1MXx7IM4srFQCgO-KTjDWL_TXqpMJU5zDzKyelrMFkc6EaMQ2KP_yBhOrh4UW-Pm7ghusox_-bV1U"
+	for j := 0; j < 10; j++ {
+		claim, err := decoder.Decode(token)
+		assert.Nil(t, err)
+		assert.NotNil(t, claim)
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	assert.Greater(t, i, 1)
```
AndrewMcCraeCA, line 35: Should this be an exact number rather than greater than 1?

jwt/encoder_options_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEncoderOptions(t *testing.T) {
+	b, err := os.ReadFile(filepath.Clean(testRSA256PrivateKey))
+	assert.Nil(t, err)
+	privKey := string(b)
+
+	i := 0
+	encoderKeyFunc := func() (string, string) {
```
AndrewMcCraeCA, line 18:
```suggestion
// This will be called each time the cache is refreshed and they we can assert i has been incremented the correct number of times below.
	encoderKeyFunc := func() (string, string) {
```

AndrewMcCraeCA (approved):

PR #362, opened 2024-03-26
cleaned up lint excludes and improved redact string

techmanmike:
Purpose: Simplify linting and log code

Changes:
- Fixed lint warning "skip dirs" -> "exclude dirs"
- Bunched all lint exceptions for x/ together
- Removed customer toSnakeCase and replaced with open source version 
- Improved redactString implementation (reduced overall length of string)


log/logger.go
```diff
@@ -2,6 +2,7 @@ package log
 
 import (
 	"github.com/rs/zerolog"
+	strcase "github.com/stoewer/go-strcase"
```
AndrewMcCraeCA, line 5: I know this adds another dependecny but I like the fact we aren't rolling our own string casing stuff in this package

AndrewMcCraeCA (approved):

log/logger.go
techmanmike: Me too - that is why I added it. I also looked for a string redact library, but couldn't find one.

PR #363, opened 2024-03-26
Log error(), fatal(), and panic() with error stack trace

techmanmike:
Purpose: Fix missing stack traces on Error, Fatal and Panic

Changes:
- Actually set the zerolog.ErrorStackMarshaler to our custom logStackTracer func
- Re-wrote the Error(), Fatal() and Panic() methods to include the Stack before the Error (as per zerolog doc)
- Upgraded all fmt.Error() calls to use "github.com/go-errors/errors" which include a stack when creating the error

`
le := l.impl.Error().Dict("error", zerolog.Dict().
	Stack().
	Err(err),
).Str("event", toSnakeCase(event))`


go.mod
```diff
@@ -259,7 +260,7 @@ require (
 	github.com/oracle/oci-go-sdk/v60 v60.0.0 // indirect
 	github.com/outcaste-io/ristretto v0.2.3 // indirect
 	github.com/packethost/packngo v0.1.1-0.20180711074735-b9cb5096f54c // indirect
-	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
+	github.com/patrickmn/go-cache v2.1.0+incompatible
```
andrewcameronsims, line 263: Does this need an extra `go mod tidy`? Direct dependencies are not usually specified in this block as far as I understand?

techmanmike: I just ran `go mod tidy` and it didn't change anything....

log/logger_test.go
```diff
@@ -121,3 +123,28 @@ func TestLoggerMethods(t *testing.T) {
...
+			Str("resource", "resource_id").
+			Int("test-number", 1),
+		).Details("logging should error tracing")
+
+	// error() adds StackTrace and WithSystemTracing() which will include "pid" "num_cpus" etc. which changes from run to run / machine to machine
+	// So we don't use a testable Example() for this
+
+	// Output:
+	//
+}
```
andrewcameronsims, line 150: Should we be making any assertions in this test? Otherwise what are we validating here? 🤔

techmanmike: Its so hard to assert a stack trace, but I was using this to debug locally...

Thoughts I what I should do here?

log/stacktrace.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/go-errors/errors"
+)
+
+// logStackTracer implements the zerolog.ErrorStackMarshaler func signature.
+func logStackTracer(err error) interface{} {
+	return stackTracer(err)
+}
+
+func stackTracer(err error) string {
+	// is it the standard google error type?
```
andrewcameronsims, line 17: Not sure what is meant by this comment?

techmanmike: it is a "go-error" or is it some other kind of error. We only know how to get the stack from a "go-error" at this stage...

log/stacktrace.go
```diff
@@ -0,0 +1,45 @@
...
+	if errors.As(err, &e) {
+		s := string(e.Stack())
+		return cleanStackTrace(s)
+	}
+
+	return cleanStackTrace(currentStack(5))
+}
+
+func currentStack(skip int) string {
+	stack := make([]uintptr, errors.MaxStackDepth)
```
andrewcameronsims, line 28: Is this an array of memory addresses? 🤔

go.mod
andrewcameronsims: All good, I must be confused 👍

log/logger_test.go
andrewcameronsims: I wonder if there's a way to strip out the non-deterministic parts of the log. Otherwise yeah I see the difficulty.

andrewcameronsims (approved):

log/stacktrace.go
techmanmike: Basically yes. Its the programme counter stack for the current go routine.

log/logger_test.go
techmanmike: Removed

log/stacktrace.go
techmanmike: Updated comment

log/stacktrace.go
```diff
@@ -0,0 +1,51 @@
...
+	buf := bytes.Buffer{}
+	for _, pc := range stack {
+		frame := errors.NewStackFrame(pc)
+		buf.WriteString(frame.String())
+	}
+
+	return buf.String()
+}
+
+func cleanStackTrace(stack string) string {
```
andrewcameronsims, line 42: We might not need to strip these characters, it sort of looks like the JSON encoder in Go does the escaping already: https://github.com/cultureamp/ca-go/commit/532c13e199ec6d8687b883c018646f949e05e9c5

What do you think, am I off base here?

andrewcameronsims: Checkout the branch and run the tests... it looks like for those characters, they are all escaped with backslash (look at the printed output).

techmanmike: So are you saying that when zerolog writes this out to JSON it will do the escaping/encoding for us automatically? And we don't need to do any cleaning here?

andrewcameronsims: Yup, that's what I reckon

andrewcameronsims (approved):

PR #369, opened 2024-04-04
Add Launchdarkly to stable packages

techmanmike:


launchdarkly/client.go
```diff
@@ -0,0 +1,222 @@
...
+		mode:               modeProxy,       // defaults to proxying requests through the LD Relay.
+		bigSegmentsEnabled: true,            // defaults to enable big segments
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	// add a logger if one wasn't added through opts
+	if c.logger == nil {
```
AndrewMcCraeCA, line 58: maybe it's worth keeping this functionality of passing in loggers
but I wonder if we should just use the log package level methods and remove the option to pass in a logger to the Client.
It will pull all the details from ENV vars in the service it's in.
and no one has user WithLogger for x/launchDarkly/flags so far?

AndrewMcCraeCA (approved):

PR #382, opened 2024-04-09
Added IsRunningViaTest to env package

techmanmike:


runtime/test_mode.go
```diff
@@ -0,0 +1,22 @@
+package runtime
+
+import (
+	"flag"
+	"os"
+	"strings"
+)
+
+// IsRunningTests returns true if the code is running in a test or is being debugged in vscode.
+func IsRunningTests() bool {
```
techmanmike, line 10: Or is this really an "env" package setting?

PR #390, opened 2024-04-15
Logger updates

techmanmike:
Purpose: Make common logging statements easier with new extensions.

Changes:
- Added WithDatadogTracing(ctx)
- Added Func field type for custom client extensions

### DIFF
--- a/log/extensions.go
+++ b/log/extensions.go
@@ -111,7 +111,7 @@ func (lf *Property) WithAuthorizationTracing(req *http.Request) *Property {
 }
 
 // WithDatadogTracing adds a "datadog" subdocument to the log that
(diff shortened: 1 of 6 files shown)


log/extensions.go
```diff
@@ -105,6 +110,73 @@ func (lf *Property) WithAuthorizationTracing(req *http.Request) *Property {
 	)
 }
 
+// WithDatadogTracing adds a "datadog" subdocument to the log that
+// includes the spans dd.trace_id and dd.span_id. If Xray is configured it also
```
davidlopezre, line 114:
```suggestion
// includes the fields dd.trace_id and dd.span_id. If Xray is configured it also
```

davidlopezre (approved):

PR #394, opened 2024-04-17
JWT: no longer panic at start up for package level  default encoder/decoder

techmanmike:
Purpose: Stop panic at startup for missing env-vars

Changes:
- Removed panics, throw errors when using instead of at startup
- Removed isTestMode
- Package level tests now load test keys (rather than rely on startup to do it)
- Renamed JWT.md to README.md and updated with the latest updates

### DIFF
--- a/jwt/README.md
+++ b/jwt/README.md
@@ -10,8 +10,6 @@ To use the package level methods `Encode` and `Decode` you MUST set these:
 - AUTH_PRIVATE_KEY = The private RSA PEM key used for Encoding a token.
 - AUTH_PRIVATE_KEY_ID = The "kid" (key_id) header to add to the token heading when Encoding.
 
-Failure to set these will result in a panic() at start up.
-
 ## Runtime Key Rotation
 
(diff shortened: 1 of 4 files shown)


AndrewMcCraeCA (approved):

jwt/README.md
```diff
@@ -64,6 +63,47 @@ You MUST set the `Issuer`, `Subject`, and `Audience` claims along with the stand
...
+	// the `exp` (Expiration Time) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4
+	ExpiresAt time.Time // default on Encode is +1 hour from now
+	// the `nbf` (Not Before) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5
+	NotBefore time.Time // default on Encode is "now"
+	// the `iat` (Issued At) claim. See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6
+	IssuedAt time.Time // default on Encode is "now"
+}
+```
+
+However, if you have wish to have customer claims then you can use the `EncodeWithCustomClaims` and `DecodeWithCustomClaims` methods.
```
kimrogers, line 93: I missed the review boat - but I think this should be "custom claims" over "customer claims"

PR #395, opened 2024-04-18
Log: errors thrown when using the logger, not at startup

techmanmike:
Purpose: Don't panic at startup, instead throw errors when used

Changes:
- Renamed LOGGER.md to README.md
- Removed isTestMode()
- Create package level DefaultLogger on use, not at startup


kimrogers (approved):
looks good, thanks Mike.

PR #396, opened 2024-04-18
Secrets: do not throw panics at startup

techmanmike:
Purpose: Don't require env-vars at startup and don't panic

Changes:
- Removed isTestMode
- Removed clients
- Renamed SECRETS.md to README.md


secrets/README.md
```diff
@@ -32,8 +30,31 @@ func BasicExamples() {
 	fmt.Printf("The answer to the secret is '%s' (err='%v')\n", answer, err)
 
 	// or if you need secrets from another region other than the one you are running in use
-	sm := secrets.NewAWSSecretsManager(ctx, "a-different-region")
+	sm, err := secrets.NewAWSSecretsManager(ctx, "a-different-region")
 	answer, err = sm.Get(ctx, "my-test-secret2")
 	fmt.Printf("The answer to the secret2 is '%s' (err='%v')\n", answer, err)
+
+	// of if you want to have a custom client that
```
kimrogers, line 37: Is this a half finished thought?

secrets/package.go
```diff
@@ -2,54 +2,34 @@ package secrets
...
+// Secrets can be mocked by clients for testing purposes.
+type Secrets interface {
+	Get(ctx context.Context, secretKey string) (string, error)
 }
 
+// DefaultAWSSecretsManager is a public *AWSSecretsManager used for package level methods.
+var DefaultAWSSecretsManager Secrets = nil
+
 // Get retrives the secret from AWS SecretsManager.
 func Get(ctx context.Context, secretKey string) (string, error) {
```
davidlopezre, line 17: Since these methods already return an error, I reckon it is better to return the error here than panicking.

techmanmike: Agree - I've fixed this in the latest commit.

secrets/README.md
techmanmike: Fixed!

kimrogers (approved):
Thanks Mike, looks good.

PR #397, opened 2024-04-18
Cipher: Do not throw panics at startup

techmanmike:
Purpose: Don't panic at startup if env-vars are missing

Changes:
- Renamed CIPHER.md to README.md
- Removed inbetween client, you now mock the DefaultKMSCipher instead
- Removed isTestMode()


kimrogers (commented):
Changes LGTM, but public api hasn't used this package so you may want more eyes on it. Otherwise happy to approve.

cipher/package.go
```diff
@@ -2,60 +2,43 @@ package cipher
...
-	}
-
-	client = newAWSKMSClient(region)
-	return NewKMSCipherWithClient(client)
-}
+var DefaultKMSCipher KMSCipher = nil
 
 // Encrypt will use env var AWS_REGION and the KMS keyId to encrypt the plainStr and return it as a base64 encoded string.
 func Encrypt(ctx context.Context, keyId string, plainStr string) (string, error) {
+	mustHaveDefaultKMSCipher()
```
davidlopezre, line 23: If the user forgot about setting AWS_REGION, what will the error look like? Would it be more clear if we returned an error straight after finding that `os.Getenv("AWS_REGION")` is "" or is the error coming out afterwards going to be explicit enough?

techmanmike: mustHaveDefaultKMSCipher now returns error if AWS_REGION not set

davidlopezre (approved):

PR #398, opened 2024-04-18
Env: removed isTestMode and defaulted env vars instead

techmanmike:
Purpose: Remove panics at startup

Changes:
- Removed isTestMode()
- Renamed ENV.md to README.md
- Defaulted env-vars to "<unknown>" to make sure engineers see and set these correctly


kimrogers (commented):
Changes LGTM, but public api hasn't used this package so you may want more eyes on it. Otherwise happy to approve.

kimrogers (approved):

PR #409, opened 2024-04-27
Log Package: return error on Details and Send instead of panic

techmanmike:
Purpose: Stop panics (very rude) and instead return errors so that clients can handle if they wish.

Changes:
- Send, Details and Detailsf all return error
- Removed mustProcess() and the panic
- Updated comments
- Fixed golint warning about unknown "skip-pkg-cache" setting

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           version: "latest"
           args: --verbose --timeout=5m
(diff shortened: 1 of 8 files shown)


davidlopezre (commented):
Approved from Public API side but would love if we could get an additional review as this could be a bit of a breaking change.

cc: @andrewcameronsims @AndrewMcCraeCA 🥺 🙏

davidlopezre (approved):

PR #410, opened 2024-04-29
Log Package: time.durations now use ISO8601 format as per logging std

techmanmike:
Purpose: 
- Compliance with the Logging Standard
-  https://cultureamp.atlassian.net/wiki/spaces/TV/pages/3114598406/Logging+Standard#Custom-fields

Changes:
- Added reference to github.com/sosodev/duration to support ISO8601 parsing and formatting
- Changed fields.Duration to covert time.duration to ISO8601 Duration string

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -159,6 +159,7 @@ linters-settings:
           - github.com/patrickmn/go-cache
           - github.com/rs/zerolog
           - github.com/segmentio
+          - github.com/sosodev/duration
           - github.com/stoewer/go-strcase
           # not sure about these below
           - github.com/cenkalti/backoff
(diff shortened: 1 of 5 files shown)


andrewcameronsims:
What is the rationale for having duration (as well as wall clock time) as iso8601? Most requests are going to last on the order of milliseconds.

techmanmike:
> What is the rationale for having duration (as well as wall clock time) as iso8601? Most requests are going to last on the order of milliseconds.

We should have this discussion on the Logging std itself - but the reason why is because its very easy to get the unit wrong on "time_taken" durations. Is it ms? or sec? etc. 

The ISO8601 was designed because - The standard provides a well-defined, unambiguous method of representing calendar dates and times in worldwide communications, especially to avoid misinterpreting numeric dates and times.

andrewcameronsims (approved):

PR #412, opened 2024-05-01
Package LOG: Updates from Go meeting

techmanmike:
Purpose:

Changes:
- removed the returned error on Debug() Info() etc.
- allowed WithDataDogTracing() etc. to be called on NewLogger() constructor (these are added to the logger for subsequent logging calls)
- added Child() which creates a new logger inheriting the properties of the parent
- can add "default_properties" by calling WithProperties() on NewLogger() constructor - these are added to the logger for subsequent logging calls
- updated README


AndrewMcCraeCA (approved):

AndrewMcCraeCA:
going to merge cause the errors from previous version are starting to make it into peoples dependency updates:
See here for example
https://github.com/cultureamp/accounts-service/pull/294

PR #419, opened 2024-05-15
Package JWT: allow clients to optionally enforce matching of aud, iss and sub claims

techmanmike:
Purpose: Allow clients to enforce matching aud, iss and sub claims

Changes:
- created DecoderParserOptions with MustMatchAudience, MustMatchIssuer and MustMatchSubject methods
- updated jwt/README.md

### DIFF
--- a/jwt/decoder_options.go
+++ b/jwt/decoder_options.go
@@ -18,3 +18,42 @@ func WithDecoderCacheExpiry(defaultExpiration, cleanupInterval time.Duration) Jw
 		decoder.cleanupInterval = cleanupInterval
 	}
 }
+
+type DecoderParserOption func(*decoderParser)
+
+type decoderParser struct {
+	expectedAud string
+	expectedIss string
+	expectedSub string
+}
+
+func newDecoderParser() *decoderParser {
+	return &decoderParser{}
+}
+
+// MustMatchAudience configures the validator to require the specified audience in
+// the `aud` claim. Validation will fail if the audience is not listed in the
+// token or the `aud` claim is missing.
+func MustMatchAudience(aud string) DecoderParserOption {
+	return func(p *decoderParser) {
+		p.expectedAud = aud
+	}
+}
+
+// MustMatchIssuer configures the validator to require the specified issuer in the
+// `iss` claim. Validation will fail if a different issuer is specified in the
+// token or the `iss` claim is missing.
+func MustMatchIssuer(iss string) DecoderParserOption {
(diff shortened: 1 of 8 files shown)


jwt/decoder_options.go
```diff
@@ -18,3 +18,43 @@ func WithDecoderCacheExpiry(defaultExpiration, cleanupInterval time.Duration) Jw
...
+func newDecoderParser() *decoderParser {
+	return &decoderParser{}
+}
+
+// MustMatchAudience configures the jwt parser to require the specified audience in
+// the `aud` claim. Validation will fail if the audience is not listed in the
+// token or the `aud` claim is missing.
+func MustMatchAudience(aud string) DecoderParserOption {
+	return func(p *decoderParser) {
+		p.expectedAud = aud
```
AndrewMcCraeCA, line 40: jwt.io seems to parse our example token as an array of strings
Screenshot 2024-05-16 at 2 02 56 PM
Do we need to cater for mutliple audiences? or would we only ever expect 1 `aud` value?

AndrewMcCraeCA (approved):
Left one comment but otherwise looks good

jwt/decoder_options.go
techmanmike: Yes - the encoded claims can specify multiple audiences, but the decoder just checks that the aud that is "itself" is in that list.  

See decoder.go line 123:

if parserOptions.expectedAud != "" {
	opts = append(opts, jwt.WithAudience(parserOptions.expectedAud))
}

That "jwt.WithAudience()" goes onto be implemented inside go-jwt as:

func (v *Validator) verifyAudience(claims Claims, cmp string, required bool) error {
	aud, err := claims.GetAudience()
	if err != nil {
		return err
	}

	if len(aud) == 0 {
		return errorIfRequired(required, "aud")
	}

	// use a var here to keep constant time compare when looping over a number of claims
	result := false

	var stringClaims string
	for _, a := range aud {
		if subtle.ConstantTimeCompare([]byte(a), []byte(cmp)) != 0 {
			result = true
		}
		stringClaims = stringClaims + a
	}

	// case where "" is sent in one or many aud claims
	if stringClaims == "" {
		return errorIfRequired(required, "aud")
	}

	return errorIfFalse(result, ErrTokenInvalidAudience)
}

Which loops through all the audiences and check that "this" audience is in that list.

Make sense?

AndrewMcCraeCA (approved):

PR #435, opened 2024-06-07
add devbox

elliotthilaire-ca:
## Purpose

Adds devbox as the means for setting up the projects locally.

## Changes

- **Add note about go@1.22.4**
- **Add .env.local to .gitignore and add loading of .env.local to .envrc**
- **Remove .envrc from .gitignore**
- **Run `devbox generate direnv`**
- **Add devbox.json**
- **Devbox add pre-commit, snyk, trufflehog**
- **Exclude devbox.json from check-json commit hook**
- **Update README.md with instructions for setting up with devbox.**
- **Remove .devcontainer config**
- **Remove .tool-versions**

## How to review this PR

 - Check out this branch and review the new [readme instructions](https://github.com/cultureamp/ca-go/tree/add-devbox?tab=readme-ov-file#installation-and-running-with-devbox).
 - After running the `devbox run setup` command ensure that you can develop on the project as expected.

## Considerations

This cleans up other means for setting up the projects i.e. asdf, devcontainers so that there's a single clear way to get up and running.

Devbox installs the requirement for the pre-commit configuration but it's up to individual developers to run `pre-commit install`. I tried adding `pre-commit install` to the automatic setup but that is too invasive because it looks like the snyk cli requires being logged into run.

We don't have a process-compose file working on this project doesn't have any running processes. Instead instructions say to run `devbox run setup` instead of `devbox services up`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .DS_Store
 /.idea
 .dccache
+.env.local
 
 # Binaries for programs and plugins
 *.exe
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /artifacts
 .DS_Store
 /.idea
-.envrc
 .dccache
 
 # Binaries for programs and plugins
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.11.0/.schema/devbox.schema.json",
+  "packages": [
+    "go@1.22",
+    "pre-commit@latest",
+    "snyk@latest",
+    "trufflehog@latest",
+  ],
+  "shell": {
+    "scripts": {
+      "setup": "make mod"
+    },
+  },
+}
--- a/devbox.json
+++ b/devbox.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.11.0/.schema/devbox.schema.json",
   "packages": [
-    "go@1.22",
+    "go@1.22", // go 1.22.4 at nixhub.io/packages/go yet. Update to go@1.22.4 when it becomes available.
     "pre-commit@latest",
     "snyk@latest",
     "trufflehog@latest",
--- a/devbox.json
+++ b/devbox.json
@@ -8,7 +8,7 @@
   ],
   "shell": {
     "scripts": {
-      "setup": "make mod"
+      "setup": "make mod",
     },
   },
 }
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -2,7 +2,6 @@
 /artifacts
 .DS_Store
 /.idea
-.envrc
```
t-monaghan, line 5: If `.envrc` was previously ignored, should we include some guidance in the readme to reflect how to handle environment secrets?

t-monaghan, line 5: If .envrc was previously ignored, should we include some guidance in the readme to reflect how to handle environment secrets?

devbox.json
```diff
@@ -0,0 +1,14 @@
...
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.11.0/.schema/devbox.schema.json",
+  "packages": [
+    "go@1.22",
+    "pre-commit@latest",
+    "snyk@latest",
+    "trufflehog@latest",
+  ],
+  "shell": {
+    "scripts": {
+      "setup": "make mod",
```
t-monaghan, line 11: Seeing as all `make` commands have phony targets can we fold these into devbox scripts?

t-monaghan (changes requested):
We discussed the issue of go modules from a devbox project existing outside of the project's isolation. My gut instinct is that this is fine, devbox itself is a go project, and if this was an issue I'm confident that the developers at Jetify would have addressed this with a default plugin. My other thought is that we should not look for a solution where we haven't had a problem yet, I'd be more inclined to solve it when we come across an issue, but it is worth some discovery.

.gitignore
elliotthilaire-ca: I've handled this here with a comment in the .envrc file. If anyone has a modified .envrc file that was private this explanation will show up when they bring in the new changes.
https://github.com/cultureamp/ca-go/pull/435/commits/5c3c51d6a37503266b5146f58f7bc97f31a4cb0e

Devs can create .env.local file that is .gitignored.  This approach allows overrides to be created while also being secure in that direnv allow is required to run when things are modified. More detail in the comment in .envrc.

devbox.json
elliotthilaire-ca: I decided against doing this.

Devbox is going to be the easiest way to set up the project but I don't want to make it difficult for anyone actively chooses to take a different route. Something like running tests is a 'project' concern rather than a 'setup project with devbox' concern.  If we move all commands into the devbox.json file then we make it harder for anyone who chooses not to use devbox.

andrewcameronsims (approved):

